PR #37, opened 2018-01-23
GitHub sign-in support, front-end edition

jpeddicord:
Substantially simpler than clearlydefined/service#56.


src/api/clearlyDefined.js
```diff
@@ -133,7 +124,7 @@ function post(url, token, payload) {
   return fetch(url, {
     headers: getHeaders(token),
     method: 'POST',
-    body: JSON.stringify(payload)
+    body: JSON.stringify(payload),
```
jeffmcaffer: is this a linting thing to _add_ trailing ','?

src/components/Header.js
```diff
@@ -10,13 +10,12 @@ import { ROUTE_ROOT } from '../utils/routingConstants'
 import { Nav, Navbar, NavItem } from 'react-bootstrap'
 import { LinkContainer, IndexLinkContainer } from 'react-router-bootstrap'
 import { filter } from 'lodash'
-import { LoginPrompt } from './'
+import { url } from '../api/clearlyDefined';
```
jeffmcaffer: ';'!  ;-)

src/api/clearlyDefined.js
jpeddicord: Habit. Did we decide on a formatter for this repo? I don't see one. `prettier`?

mynameistechno: +1 for `prettier`!

jpeddicord: Agreed. I'll quick-fix these format issues you've noted in this review, but let's save the format pass for another commit.

PR #56, opened 2018-02-23
First stab at remembering FilterBar values

### LINKED ISSUE #35: fill in filter bar when navigating back
In the /inspect page, if you select a component then navigate away (say the curate page) and then come back, the filter bar is not filled in but the data for component you were inspecting is still there/re-populated.

We should use the `selected` prop on the `Typeahead` component in the `FilterBar` to show the current value to the user.

note that this may be relevant for other pages as well.

mynameistechno:
This addresses some of the issues with the FilterBar remembering the selected value. There are still a few things to work out though, but submitting this PR now since I'm out for a couple days.
* I took a stab at keeping the URL up to date with the selected definition but this introduced other problems so I pulled it out for now (i.e. I made progress leveraging `gotoValue` so that the URL also matched what the user selected in the FilterBar, etc but ran into some other funky issues with routing). **We should look to globally solve keeping state and url history in sync.** We may want to look into using `react-router-redux` or something. Ideally we just manage state, and routing is updated automatically (or vice versa - when you navigate somewhere, state gets updated). This needs some more research.
* Curate page: when you select something, browse away and come back, the filterbar value is filled in but the curation is gone. In `PageCurate.componentDidMount` , replacing `this.filterChanged` to `this.handleNewSpec` fixes it but then this breaks deep linking (i.e. when you have a curation in the URL path and refresh).

Fixes #35

### DIFF
--- a/src/components/FilterBar.js
+++ b/src/components/FilterBar.js
@@ -20,6 +20,9 @@ export default class FilterBar extends Component {
     super(props)
     this.onChange = this.onChange.bind(this)
     this.filter = this.filter.bind(this)
+    this.state = {
+      value: props.value ? [props.value] : null
+    };
   }
 
   onChange(values) {
@@ -50,6 +53,7 @@ export default class FilterBar extends Component {
         defaultInputValue={defaultValue || ''}
         filterBy={this.filter}
         labelKey='path'
+        selected={this.state.value}
       />
     )
   }
--- a/src/components/FilterBar.js
+++ b/src/components/FilterBar.js
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.
+// Copyright (c) Microsoft Corporation and others.
 // SPDX-License-Identifier: MIT
 
 import React, { Component } from 'react'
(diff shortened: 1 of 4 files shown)


src/components/FilterBar.js
```diff
@@ -47,9 +53,9 @@ export default class FilterBar extends Component {
         options={options.transformedList}
         isLoading={options.isFetching}
         clearButton
-        defaultInputValue={defaultValue || ''}
```
mynameistechno: FYI I removed this because it conflicts with `selected`:

image

There's a `defaultSelected` that can be used instead, but this caused other issues. It worked best by just passing in `path` directly into `selected` if it's set.

mynameistechno:
Regarding updating the URL:
When you arrive on the Inspect Page by clicking on the Dig deeper icon on Browse, the URL includes that component. If you then proceed to select another component via filtering, the URL is not updated so if you copy-paste/bookmark the URL at that point, it’s incorrect. So to avoid that I decided to always represent the selected item in the URL. Anyway that was my thinking, which then led me to think we need a better way to manage the url/history across the app

mynameistechno:
Rebased this. @jeffmcaffer any feedback? I added a couple notes above to help explain a couple decisions. I think we should revisit some of this, but it works for now. Long term, I feel like dispatching a single action should be enough to update (and navigate the user to) the new URL and set the filterValue. E.g. on the Browse page, if they hit the Inspect button, one action should trigger a couple reducers that update filterValue and the URL. I need to look into `react-router-redux`, which claims to keep react-router and redux in sync.

mynameistechno:
Closing this PR. Going to submit another one that focuses just on maintaining the `Filterbar` selection. I've created a separate ticket to track the other bits #112

PR #63, opened 2018-02-27
Adds a copy-url-button component

mynameistechno:
This PR:
* Adds a copy-url-button component that leverages `react-copy-to-clipboard`
* Adds the button to components list, inspect, and curate pages (not sure we want them on all, but can easily change)
* A couple UX clean up items (just experimenting, I can revert/change):
  * Groups the icon buttons in the component list
  * Moves the badge over the left so it doesn't shift the buttons to the left after it renders (this is temporary, long term we can do better)
  * Adds bootstrap's "well" styling around the filter boxes

image

On the Inspect/Curate pages I tried a more subtle "button link" style floated to the right. I didn't want the user to confuse it with a search button (but it looks good as a button and I'm open to changing it or removing altogether):

image

### DIFF
--- a/src/components/CopyUrlButton.js
+++ b/src/components/CopyUrlButton.js
@@ -10,7 +10,17 @@ import FontAwesome from 'react-fontawesome'
 export default class CopyUrlButton extends Component {
 
   static propTypes = {
-    path: PropTypes.string
+    path: PropTypes.string,
+    bsStyle: PropTypes.string,
+    className: PropTypes.string,
+    route: PropTypes.string
+  }
+
+  static defaultProps = {
+    path: '',
+    bsStyle: 'link',
+    className: '',
+    route: ''
   }
 
   constructor(props) {
@@ -31,11 +41,11 @@ export default class CopyUrlButton extends Component {
   onCopy() {
     this.state.timeoutId && clearTimeout(this.state.timeoutId)
     const timeoutId = setTimeout(this.didCopy, 2000)
-    this.setState({...this.state, copied: true, timeoutId})
+    this.setState({ ...this.state, copied: true, timeoutId })
   }
 
   didCopy() {
-    this.setState({...this.state, copied: false, timeoutId: null})
+    this.setState({ ...this.state, copied: false, timeoutId: null })
   }
 
   renderUrl() {
@@ -50,7 +60,7 @@ export default class CopyUrlButton extends Component {
   }
 
   render() {
-    const {path, bsStyle = 'link', className} = this.props
+    const {path, bsStyle, className} = this.props
     const isDisabled = !Boolean(path)
 
     return (
--- a/src/components/CopyUrlButton.js
+++ b/src/components/CopyUrlButton.js
@@ -1,0 +1,64 @@
+// Copyright (c) 2018, The Linux Foundation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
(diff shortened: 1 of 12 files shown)


src/components/CopyUrlButton.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  renderTooltip() {
+    return (
+      <Tooltip id="tooltip">{this.state.copied ? "Copied!" : "Copy URL to clipboard"}</Tooltip>
+    )
+  }
+
+  render() {
+    const {path, bsStyle = 'link', className} = this.props
```
jeffmcaffer: suggest adding the `bsStyle` default value as a default prop to make it clear to component users.

src/components/CopyUrlButton.js
```diff
@@ -0,0 +1,66 @@
...
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { Button, Tooltip, OverlayTrigger } from 'react-bootstrap'
+import { CopyToClipboard } from 'react-copy-to-clipboard'
+import FontAwesome from 'react-fontawesome'
+
+export default class CopyUrlButton extends Component {
+
+  static propTypes = {
+    path: PropTypes.string
```
jeffmcaffer: should have bsStyle and className as declared props (with defaults ideally)

src/components/CopyUrlButton.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  onClick(event) {
+    event.stopPropagation()
+  }
+
+  onCopy() {
+    this.state.timeoutId && clearTimeout(this.state.timeoutId)
+    const timeoutId = setTimeout(this.didCopy, 2000)
+    this.setState({...this.state, copied: true, timeoutId})
```
jeffmcaffer: formatting nit wrt space after { and before } for object literals.

src/components/PageComponentDetails.js
```diff
@@ -118,10 +118,13 @@ class PageInspect extends Component {
     const { filterOptions, filterValue, definition, curation, harvest } = this.props
     return (
       <Grid className='main-container'>
-        <Row className="show-grid spacer">
-          <Col md={10} mdOffset={1}>
+        <Row className="show-grid well">
```
jeffmcaffer: not keen on this being a well. The design is meant to be super clean without boxes, borders, grid lines, ...  Is this needed?

src/components/CopyUrlButton.js
mynameistechno: Prettier wherefore art thou 😜

src/components/CopyUrlButton.js
mynameistechno: Sounds good, will define and add defaults for all props.

src/components/PageComponentDetails.js
mynameistechno: No I can remove, it's just that `spacer` had issues. I'll just add more margin between navbar and content to achieve same effect as `spacer`.

mynameistechno:
Feedback addressed @jeffmcaffer. Also, I removed the `well` class and did some minor cleanup in the components list. The source line/sha was causing wrapping on the buttons on the right, so I added ellipsis text overflow support for the source value so the layout handles large source values more gracefully (and a couple other css fixes):

Before:
image

After:
image

The panel/list expand below needs some work, but that can be tackled separately. E.g. if browser width is narrowed a bit:
image

Created #74 to that effect.

We should prob merge #68 first so I can make sure the filterbar renders well with the facets.

PR #64, opened 2018-02-27
Remove source filter action buttons from page

craigez:
Signed-off-by: Craig Northway <cnorthway@codeaurora.org>


src/components/PageComponents.js
```diff
@@ -14,14 +14,13 @@ class PageComponents extends Component {
 
   constructor(props) {
     super(props)
-    this.state = { activeProvider: 'github' }
+    this.state = { }
```
jeffmcaffer: you can just delete setting the state unless it is used for something. No technical harm in having it but points people to think that the component actually has/develops meaningful state.

PR #68, opened 2018-02-27
Add faceting on browse of components

### LINKED ISSUE #62: add filtering by facet on Browse
We now have the ability to strip out facets from the data we have. This will be particularly useful in the Browse details pane and for the badge calculations. 

A user should be able to multiselect any of the 6 facets (core, data, dev, docs, examples, tests) and then all they see in the Browse view only includes that data.

craigez:
Resolves #62 

Signed-off-by: Craig Northway <cnorthway@codeaurora.org>


craigez:
This will clash with #64 I think it also achieves that clean up, so we can probably close #64

src/components/FacetSelect.js
```diff
@@ -0,0 +1,59 @@
...
+import 'react-select/dist/react-select.css';
+
+const options = [
+  { value: 'core', label: 'Core'},
+  { value: 'data', label: 'Data'},
+  { value: 'dev', label: 'Dev'},
+  { value: 'docs', label: 'Docs'},
+  { value: 'examples', label: 'Examples'},
+  { value: 'tests', label: 'Tests'}
+]
```
craigez: I don't know if we want this in here or in the PageComponent to generalize this further.

craigez:
screen shot 2018-02-27 at 4 00 55 pm

src/components/ComponentList.js
```diff
@@ -175,6 +178,7 @@ export default class ComponentList extends React.Component {
     let expressions = []
     let declared = []
 
+    console.log("Facets: " + JSON.stringify(facets))
```
jeffmcaffer: hmmm, let's remove this

src/components/ComponentList.js
```diff
@@ -326,8 +330,9 @@ export default class ComponentList extends React.Component {
   }
 
   render() {
-    const { loadMoreRows, listHeight, noRowsRenderer, list, fetchingRenderer } = this.props
+    const { loadMoreRows, listHeight, noRowsRenderer, list, fetchingRenderer, facetList } = this.props
     const { sortOrder, contentSeq } = this.state
+    console.log("Rendering with facetList:" + JSON.stringify(facetList))
```
jeffmcaffer: and this and other console.log calls

src/components/FacetSelect.js
```diff
@@ -0,0 +1,59 @@
...
+    this.setState({value})
+    console.log('Current state:', JSON.stringify(this.state))
+    const { changeHandler } = this.props;
+    changeHandler && changeHandler(value)
+  }
+
+  render () {
+    const { value } = this.state;
+    return (
+      <div className="section">
```
jeffmcaffer: curious about this class. What effect you are looking for? I don't see a `section` class anywhere. What happens if you just delete the div altogether?

src/components/PageComponents.js
```diff
@@ -60,15 +63,34 @@ class PageComponents extends Component {
     this.props.dispatch(uiBrowseUpdateList({ update: component, value: newComponent }))
   }
 
+  facetChange(value) {
+    const activeFacets = [];
+    value && value.forEach(facet => activeFacets.push(facet['value']))
```
jeffmcaffer: consider 
```javascript
const activeFacets = (value || []).map(facet => facet.value)
```

src/components/PageComponents.js
```diff
@@ -60,15 +63,34 @@ class PageComponents extends Component {
...
+    this.setState({ ...this.state, activeFacets})
+    console.log('Current state in Page Component:', JSON.stringify(this.state))
+  }
+
+  renderActionButton() {
+    return (<Button className='pull-right' bsStyle='success' onClick={this.filterHandler}>Filter</Button>)
```
jeffmcaffer: do we still need this Filter button? should we just trigger the filterHandle when the active facet list is changed?

src/components/PageComponents.js
```diff
@@ -60,15 +63,34 @@ class PageComponents extends Component {
...
   render() {
     const { components, filterOptions, definitions, token } = this.props
+    const { activeFacets } = this.state
     return (
       <Grid className='main-container'>
         <Row className='show-grid spacer'>
+          <Col md={4}>
+            <FacetSelect 
+              name="facets"
+              changeHandler={this.facetChange}
```
jeffmcaffer: typical name would be `onChange`. Not a rule, just seems to common "event style" naming

src/components/FacetSelect.js
jeffmcaffer: Feels fine here... it could be another prop but at least for now, Facets and the list are pretty hard concepts in the system. We'll see if we need this same sort of control and want generalize it to just a "multiselector" or something.

src/components/ComponentList.js
craigez: yeah, committed without cleaning up my debugging

src/components/FacetSelect.js
craigez: I added this from an example when I was debugging visual issues with the component, didn't even check the styles.

craigez:
I'll clean it up sometime today.

src/components/PageComponents.js
craigez: That handler does nothing at the moment anyway, so removing it.

craigez:
@jeffmcaffer feedback resolved

PR #71, opened 2018-02-28
Add notification alerts

### LINKED ISSUE #14: Add feedback for operation progress/success/failure
Need Alerts (or some such) saying that an operation is in progress and that it has finished (success/failure).

mynameistechno:
This adds an easy way to trigger notification alerts in the app. It's essentially a container that manages state for the [React Bootstrap Notifier](https://chadly.github.io/react-bs-notifier/). Example: to trigger a notification:

```js
import { uiNotificationNew } from '../actions/ui'
this.props.dispatch(uiNotificationNew({ type: 'info', message: 'Harvesting started.', timeout: 5000 }))
```

For all available options see https://chadly.github.io/react-bs-notifier/ 

To demonstrate usage the Harvest button triggers an "info" alert that harvesting has started. Subsequent PRs can dispatch notifications to the success/error handlers for the harvester.

image

Fixes #14

### DIFF
--- a/src/components/NotificationList.js
+++ b/src/components/NotificationList.js
@@ -10,22 +10,32 @@ import { uiNotificationDelete } from '../actions/ui'
 class NotificationList extends Component {
 
   static propTypes = {
-    notifications: PropTypes.array
+    notifications: PropTypes.array,
+    position: PropTypes.string
   }
 
   static defaultProps = {
-    notifications: []
+    notifications: [],
+    position: 'top-right'
+  }
+
+  constructor(props) {
+    super(props)
+    this.onDismiss = this.onDismiss.bind(this)
+  }
+
+  onDismiss(message) {
+    this.props.dispatch(uiNotificationDelete(message))
   }
 
   render() {
-    const {dispatch} = this.props
+    const { position, notifications } = this.props
     return (
       <div>
         <AlertList
-          position="top-right"
-          alerts={this.props.notifications}
-          timeout={4000}
-          onDismiss={message => dispatch(uiNotificationDelete(message))}
+          position={ position }
+          alerts={ notifications }
+          onDismiss={ this.onDismiss }
           />
       </div>
     )
--- a/src/components/NotificationList.js
+++ b/src/components/NotificationList.js
@@ -1,0 +1,39 @@
+// Copyright (c) 2017, The Linux Foundation. All rights reserved.
+// SPDX-License-Identifier: MIT
+
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+import { AlertList } from 'react-bs-notifier'
+import { uiNotificationDelete } from '../actions/ui'
+
+class NotificationList extends Component {
+
+  static propTypes = {
+    notifications: PropTypes.array
+  }
+
+  static defaultProps = {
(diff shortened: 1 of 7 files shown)


src/components/NotificationList.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  render() {
+    const {dispatch} = this.props
+    return (
+      <div>
+        <AlertList
+          position="top-right"
+          alerts={this.props.notifications}
+          timeout={4000}
```
jeffmcaffer: make this a prop? with a default?

src/components/NotificationList.js
```diff
@@ -0,0 +1,39 @@
...
+          onDismiss={message => dispatch(uiNotificationDelete(message))}
```
jeffmcaffer: feels funny for this component to reach out for an action. perhaps pass in the dismiss action as a prop

src/components/PageHarvest.js
```diff
@@ -28,6 +28,7 @@ class PageHarvest extends Component {
 
   harvestHandler(spec) {
     const { dispatch, token, queue } = this.props
+    dispatch(uiNotificationNew({ type: 'info', message: 'Harvesting started.', timeout: 5000 }))
```
jeffmcaffer: what is this timeout? vs the AlertList timeout (which is shorter)

jeffmcaffer: Not required up front (this is a huge improvement) but, ...  Any thought on how to get this going away when the response comes (as opposed to just timing out), and reporting errors?

src/components/NotificationList.js
mynameistechno: Sure for consistency. But if you wanted to change this for a specific alert you wouldn't use a prop, you pass it in the object to the action:
```js
dispatch(uiNotificationNew({ type: 'info', message: 'foo', timeout: 15000 }))
```

mynameistechno: Actually, I should remove it the default! The timeout specifies when it should auto-hide. With the default in place you can never create a notification that doesn't auto-hide (eg. a failure you want the user to acknowledge).

src/components/NotificationList.js
mynameistechno: That's fine, I was just following some example online for something similar.

src/components/PageHarvest.js
mynameistechno: The timeout specifies when it should auto-hide.  I've removed the AlertList default.

mynameistechno: The notifications system works well with multiple alerts. So the *easiest* thing would be to just add new error/complete alerts (they appear below the harvesting one, which would then go away, and the error one below it slides up in it's place):
image

Having said that, each notification gets an id that is a part of the action's return object. So all you need to do in those cases is store that id. Then you can delete or update that message in the redux state. So maybe in the harvester reducer, look our for the `UI_NOTIFICATION_NEW` action where the message is 'Harvesting started' (we may want to store the messages as constants), then store that message ID as part of the current harvest request. When the harvest is complete, check if the id is still in the notifications list, if so you can update it or delete it.

mynameistechno:
@jeffmcaffer feedback addressed. LMK if you have any more questions.

PR #113, opened 2018-03-18
Remember selection in FilterBar

### LINKED ISSUE #35: fill in filter bar when navigating back

mynameistechno:
These couple of updates seemed to be sufficient. Fixes #35 

Signed-off-by: Mark Matyas <mmatyas@codeaurora.org>

### DIFF
--- a/src/components/FilterBar.js
+++ b/src/components/FilterBar.js
@@ -36,7 +36,7 @@ export default class FilterBar extends Component {
   }
 
   render() {
-    const { options, defaultValue } = this.props
+    const { options, defaultValue, value } = this.props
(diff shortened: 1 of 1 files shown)


mynameistechno:
> LGTM. I'll let you merge at your leisure.

Thanks @jpeddicord!

PR #116, opened 2018-03-19
Move to Font Awesome 5

### LINKED ISSUE #86: move to fontawesome 5 to get Discord
Fontawesome 5 has a Discord icon we need for the footer.

Moving likely means moving off the current react-fontawesome component as they only work with 4.* it seems.

OTOH, https://www.npmjs.com/package/@fortawesome/react-fontawesome works with FA5 but uses a different code pattern. sigh.

Good news is though that FA5 has a mess of interesting JS transforms etc that this other component enables.  So we can do some fun things if we move.

If we add this, the discord icon will automatically show up in the footer as the entry has already been made.

jpeddicord:
Fixes #86 

Went the React-less route since it doesn't require explicit icon importing. Not much work to switch back if we want.


jeffmcaffer:
Nice

PR #122, opened 2018-04-13
refactor for inline edit

jeffmcaffer:
DO NOT MERGE

A prototype of adding inline editing to the current Browse UI. This uses the inline edit function from @jpeddicord  and merges in the effect of the EditableComponent by factoring out DefinitionEntry from the original ComponentList

Currently only the Declared license and the Release Date are editable.


src/components/DefinitionEntry.js
```diff
@@ -0,0 +1,309 @@
...
+
+  static defaultProps = {}
+
+  inspectComponent(component, event) {
+    event.stopPropagation()
+    const action = this.props.onInspect
+    action && action(component)
+  }
+
+  curateComponent(component, event) {
```
jpeddicord: FYI: you can avoid the need to `.bind` by writing this as an arrow function: `curateComponent = (component, event) => { ...`

jpeddicord:
Branching off of this to hack on.

src/components/DefinitionEntry.js
jpeddicord: Fair. I'm generally more partial to #a. The main downsides I'm aware of with arrow-functions-everywhere:

1. Sometimes more difficult to test
2. Can't subclass & override arrow functions (easily)

With React components it often isn't an issue. It's a style preference for the most part, sure. My mental workflow is something like this:

* I see an arrow function -> ok, it's some class method or callback. "normal."
* I see a traditional class method -> this might be a subclass of something, or something static.
* I see `.bind` -> oh, there's probably something special going on somewhere.

I can get behind `.bind` if that's the chosen style; I just need to flip the way I think about it.

jeffmcaffer: You and I have different definitions of "normal" ;-)

There must be an execution/memory cost as a function/closure is created containing `this` for the arrow functions for each instance of the class. Poked around and found the following that makes class property arrow functions quite unattractive.

https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

PR #131, opened 2018-05-24
created new request to handle multi component curations

Teju-Manchenella:


src/components/PageDefinitions.js
```diff
@@ -87,10 +87,16 @@ class PageDefinitions extends Component {
       const coord = EntitySpec.asRevisionless(component)
       const patch = find(result, p => { return EntitySpec.isEquivalent(p.coordinates, coord) })
       const revisionNumber = component.revision
+      const patchChanges = Object.getOwnPropertyNames(component.changes).reduce((result, change) => {
```
Teju-Manchenella: I updated the buildSpec function because in the version we worked on together the object was being created as follows (this is a specific example but similar patterns would be seen with other changes)
```javascript
revision:
    revisionNumber:
        described.facets: (facet change)
        described.releaseDate: (releaseDate change)
```

What we want instead is
```javascript
 revision:
    revisionNumber:
        described:
            facets: (facet change)
            releaseDate: (releaseDate change)
```

Due to this fact I first reduced the changes. This was something that you had mentioned earlier and seems to help out our current situation. Then I assigned them as [revisionNumber] : reducedChanges.

PR #134, opened 2018-06-06
File facet tagging component

jpeddicord:
This implements a file tagging/picking component we discussed during the weekly meeting.

If you run this, you'll find a little demo at http://localhost:3000/file-tagger-test to play around with. That can be deleted at any time, either for this PR or later. The key component to use is `FileTagger` -- it takes a list of filenames, a map of file paths to facets (`facetMap`), and a callback. The callback is a change notification for `facetMap`.

This component does its best to elide duplicate information; e.g. if a child has the same facet as its parent it will only list the parent.

Here's a little video in action if you don't want to fire this all up: https://streamable.com/2zjpa

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "deep-diff": "^0.3.8",
     "lodash": "^4.17.4",
     "moment": "^2.20.1",
+    "prop-types": "^15.6.1",
     "react": "^16.2.0",
     "react-bootstrap": "^0.31.5",
     "react-bootstrap-typeahead": "^2.3.0",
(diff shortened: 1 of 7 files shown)


jeffmcaffer:
Nice. I like the approach especially as we have a small, limited set of tags. Would be interesting as well if the same thing could be coupled with values like the copyright holders and the license data for the file.

jeffmcaffer:
@jpeddicord can we close this one? there is some separate work going on to use the React table tree and support this scenario.

jpeddicord:
Yeah, I think we can close this out.

Issue #140, opened 2018-06-11
batch the adding of coordinates to the browse list on file drop


jeffmcaffer:
when a file is dropped on the Browse list we need to figure out all the coordinates and then add them all at once rather than one at a time. 

From there also ensure that the service is called once with the batch of coords to get the definitions rather than once for each.

jeffmcaffer:
done

Issue #139, opened 2018-06-11
update the browse list entry fetching to cache better


jeffmcaffer:
seems like if there is a long browse list, scrolling causes the newly revealed elements to be dynamically fetched. This is a good thing but it also appears that this happens repeatedly as the user scrolls up and down. The fetched definitions should be cached and checked so we done DOS ourselves with our own UI.

jeffmcaffer:
done

Issue #138, opened 2018-06-11
cache the clearlydefined badges so as to not abuse shields.io


jeffmcaffer:
Do this in the website app with a simple memory cache

rspier:
This is an actual, not theoretical problem.  @jeffmcaffer reports that users are seeing broken images, likely due to rate limiting.

gauntface:
Yeah - cloudflare is returning 429 response (too many requests)

gauntface:
We can add a cache control header to the badges on api.clearlydefined.io/badges/* but I don't believe that's the actual issue here. 

Cloudflare is actively rate limiting the value.

<img width="483" alt="screen shot 2018-07-16 at 11 54 40 am" src="https://user-images.githubusercontent.com/139760/42777507-2722aff4-88ef-11e8-865e-24bdb0ea9c43.png">

The best approach would be to add a rule to cloudflare to alter this rate limit (for api.cloudflare.com/* OR api.cloudflare.com/badges/*)

Worth noting that once the badges hit the rate limit, it starts to rate limit the /definitions  calls as well (before hitting the rate limit they didn't seem to contribute to the rate limit)

Issue #137, opened 2018-06-11
Add support for dropping package-lock.json


jeffmcaffer:
Allow users to drop package-lock.json (and perhaps packages.json) files on the Browse dropzone.

jeffmcaffer:
first cut implementation done

PR #142, opened 2018-06-11
add support for dropping package-lock.json files

jeffmcaffer:
Implement the feature in #137 .

Note that this should not be merged/used until we address the caching and batching work in #138, #139 and #140. Otherwise we'll DOS shields.io and ourselves.

### DIFF
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -46,27 +46,53 @@ class PageDefinitions extends Component {
     dispatch(uiBrowseUpdateList({ add: component }))
   }
 
-  onDrop(acceptedFile, rejectedFiles) {
+  onDrop(acceptedFiles, rejectedFiles) {
     const { dispatch, token, definitions } = this.props
-    acceptedFile.forEach(file => {
-      const reader = new FileReader();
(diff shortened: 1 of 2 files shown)


src/components/PageDefinitions.js
```diff
@@ -46,29 +46,55 @@ class PageDefinitions extends Component {
...
+      if (object.coordinates) return object
+      return 'No component coordinates found'
+    } catch (e) {
+      return e.message
+    }
+  }
+
+  loadPackageLockFile(dependencies) {
+    const coordinates = []
+    for (const dependency in dependencies) {
```
Teju-Manchenella: Change the for loop to a forEach to make more readable and match style of remaining project

src/components/PageDefinitions.js
```diff
@@ -46,29 +46,55 @@ class PageDefinitions extends Component {
...
+      let [namespace, name] = dependency.split('/')
+      if (!name) {
+        name = namespace
+        namespace = null
+      }
+      coordinates.push({ type: 'npm', provider: 'npmjs', namespace, name, revision: dependencies[dependency].version })
```
Teju-Manchenella: Just a question about package-lock.json files. The coordinate that is created on line 93 has 'npm' and 'npmjs' hardcoded as their respective type and provider. Are package-lock.json files only found in npm projects?

PR #143, opened 2018-06-11
Single UI Render

Teju-Manchenella:
UI only rendered once when loaded from file

### DIFF
--- a/src/reducers/listReducer.js
+++ b/src/reducers/listReducer.js
@@ -25,6 +25,14 @@ const add = (list, item, comparator = null) => {
   return list && !_.find(list, test) ? [...list, item] : list
 }
 
+const addAll = (list, items, comparator = null) => {
+  items.forEach(item => {
+    const test = comparator ? element => comparator(element, item) : element => element === item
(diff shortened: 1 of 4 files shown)


src/reducers/listReducer.js
```diff
@@ -25,6 +25,15 @@ const add = (list, item, comparator = null) => {
   return list && !_.find(list, test) ? [...list, item] : list
 }
 
+const addAll = (list, items, comparator = null) => {
+  let newList = list
+  items.forEach(item => {
+    const test = comparator ? element => comparator(element, item) : element => element === item
+    newList = newList && !_.find(newList, test) ? [...newList, item] : newList
```
jeffmcaffer: Hmmm, likely should use a lodash union-ish function. this iwll make 00s of copies of the array (`[...newList]` makes a copy of the array).

jeffmcaffer:
@Teju-Manchenella you want to roll these changes into the other `listReducer` changes you are doing? 

Also note the comment I made in `addAll`. In particular, see https://lodash.com/docs/4.17.10#unionWith

PR #148, opened 2018-06-14
Various UI changes

Teju-Manchenella:
Various UI changes were done:

- Component inline edit fields are wider.

- Facets are not editable anymore.

- The facet filter in the browse page is removed.

- The search bar in the browse page is wider.

- Link to component website, specific component revision website, and source website. Links only supported for git and npm types.


src/components/InlineEditor.js
```diff
@@ -51,9 +51,12 @@ export default class InlineEditor extends React.Component {
     const changed = initialValue !== value
     if (!editing)
       return (
-        <span className={`editable-field ${value ? (changed ? 'bg-info' : '') : 'placeholder-text'}`} onClick={() => this.setState({ editing: true })}>
+        <span
```
jeffmcaffer: I find it odd that this is reformatted. I wonder if your prettier is somehow configured differently...

src/components/InlineEditor.js
```diff
@@ -77,6 +80,6 @@ export default class InlineEditor extends React.Component {
   }
 
   editors = {
-    text: value => <input type="text" defaultValue={value} />
+    text: value => <input size="45" type="text" defaultValue={value} />
```
jeffmcaffer: Feels like the size value should be in CSS somewhere/somehow

src/components/DefinitionEntry.js
```diff
@@ -71,16 +71,28 @@ export default class DefinitionEntry extends React.Component {
 
   renderHeadline(definition) {
     const { namespace, name, revision } = definition.coordinates
+    const locationUrl = this.getLocationURL(definition)
```
jeffmcaffer: use consistent casing on URL. suggest using Url. I vaguely think that is what we use elsewhere but a quick search in the website and service code will reveal the truth.

src/components/DefinitionEntry.js
```diff
@@ -71,16 +71,28 @@ export default class DefinitionEntry extends React.Component {
...
+    let revisionText = (
+      <span>
+        &nbsp;&nbsp;&nbsp;
+        <a href={revisionUrl} target="_blank">
+          {revision}
+        </a>
+      </span>
+    )
+
     const location = get(definition, 'described.sourceLocation')
```
jeffmcaffer: note that `location` here is the **source** location, not the location of the component itself. We should likely just stop showing the source location in the two line view

src/components/DefinitionEntry.js
```diff
@@ -71,16 +71,28 @@ export default class DefinitionEntry extends React.Component {
...
-    if (!location || (definition.coordinates.provider === location.provider && revision === location.revision))
+    if (location && (definition.coordinates.provider === location.provider && revision === location.revision))
```
jeffmcaffer: suspect that this should change/removed in light of the above comment

src/components/DefinitionEntry.js
```diff
@@ -91,6 +103,26 @@ export default class DefinitionEntry extends React.Component {
     return licenseExpression ? <span>{licenseExpression}</span> : <span>&nbsp;</span>
   }
 
+  getRevisionUrl(definition) {
+    const coordinate = get(definition, 'coordinates')
+    switch (coordinate.provider) {
+      case 'github':
+        return ''
```
jeffmcaffer: github has revisions. they are commit hashes and you can get to the state of the repo for a given commit.

src/components/DefinitionEntry.js
```diff
@@ -91,6 +103,26 @@ export default class DefinitionEntry extends React.Component {
...
+      case 'npmjs':
+        return `${this.getLocationURL(definition)}/v/${coordinate.revision}`
+    }
+  }
+
+  getLocationURL(definition) {
```
jeffmcaffer: suggest a different name. "location" is too generic and redundant with "url". Perhaps `getComponentUrl` vs `getRevisionUrl`

src/components/DefinitionEntry.js
```diff
@@ -71,16 +71,28 @@ export default class DefinitionEntry extends React.Component {
...
     const { namespace, name, revision } = definition.coordinates
+    const locationUrl = this.getLocationURL(definition)
+    const revisionUrl = this.getRevisionUrl(definition)
     const namespaceText = namespace ? namespace + '/' : ''
     const sourceUrl = this.getSourceUrl(definition)
-    let revisionText = <span>&nbsp;&nbsp;&nbsp;{revision}</span>
+    let revisionText = (
+      <span>
+        &nbsp;&nbsp;&nbsp;
+        <a href={revisionUrl} target="_blank">
```
jeffmcaffer: what happens if there is no `revisionUrl`

src/components/DefinitionEntry.js
```diff
@@ -71,16 +71,28 @@ export default class DefinitionEntry extends React.Component {
...
+
     const location = get(definition, 'described.sourceLocation')
-    if (!location || (definition.coordinates.provider === location.provider && revision === location.revision))
+    if (location && (definition.coordinates.provider === location.provider && revision === location.revision))
       revisionText = ''
     return (
       <span>
-        {namespaceText}
-        {name}
+        <a href={locationUrl} target="_blank">
```
jeffmcaffer: what happens when there is no `locationUrl`?

src/components/DefinitionEntry.js
```diff
@@ -91,6 +103,26 @@ export default class DefinitionEntry extends React.Component {
...
   }
 
+  getRevisionUrl(definition) {
+    const coordinate = get(definition, 'coordinates')
+    switch (coordinate.provider) {
+      case 'github':
+        return ''
+      case 'npmjs':
+        return `${this.getLocationURL(definition)}/v/${coordinate.revision}`
+    }
```
jeffmcaffer: need to fill in the other providers

src/components/DefinitionEntry.js
```diff
@@ -91,6 +103,26 @@ export default class DefinitionEntry extends React.Component {
...
+  }
+
+  getLocationURL(definition) {
+    const coordinate = get(definition, 'coordinates')
+    switch (coordinate.provider) {
+      case 'github':
+        return `https://${coordinate.provider}.com/${coordinate.namespace}/${coordinate.name}`
```
jeffmcaffer: strictly speaking the `provider` value is NOT the name of the website. It is a symbolic label that just happens to frequently (but not always) match up. Just hardcode github.com and npmjs.com etc here

src/components/DefinitionEntry.js
```diff
@@ -91,6 +103,26 @@ export default class DefinitionEntry extends React.Component {
...
+      case 'npmjs':
+        return `https://${coordinate.provider}.com/package/${coordinate.name}`
+    }
+  }
```
jeffmcaffer: is a default case needed?

src/components/DefinitionEntry.js
```diff
@@ -244,14 +276,7 @@ export default class DefinitionEntry extends React.Component {
...
             <Col md={10}>
-              <InlineEditor
-                type="text"
-                initialValue={this.printArray(initialFacets)}
-                value={this.printArray(this.getValue('described.facets') || initialFacets)}
-                onChange={this.fieldChange('described.facets', isEqual, this.parseArray)}
-                validator={value => true}
-                placeholder={'Facets'}
-              />
+              <p className="list-singleLine"> &nbsp;&nbsp;&nbsp;&nbsp;{this.printArray(initialFacets)}</p>
```
jeffmcaffer: 4 &nbsp; ? does that line up well? In the grander scheme of things we should be using CSS etc for layout. OK to be lazy right now as long as it looks reasonable in our default setup.

src/components/InlineEditor.js
Teju-Manchenella: There are some other places where prettier does similar things with my code when I save it. I am not sure exactly how it is supposed be configured and formatted.

src/components/DefinitionEntry.js
Teju-Manchenella: Url is the correct casing. I looked through website and service code.

src/components/DefinitionEntry.js
Teju-Manchenella: I have removed source location from the two line view.

src/components/DefinitionEntry.js
Teju-Manchenella: The code was structured so source url was shown in the the TwoLineEntry. Due to this fact I decided not to have a github revision url because it would be the exact same as the source url. However, I removed the source url from the TwoLineEntry, so github revision now has the proper url template.

src/components/DefinitionEntry.js
Teju-Manchenella: I attempted to add more providers. This was the best I could do with the data I had access to. I have urls for providers github, npmjs, nuget, and mavencentral.

src/components/DefinitionEntry.js
Teju-Manchenella: If there is no locationUrl the namespace and name is shown without a hyperlink. Along with that sometimes there isn't a namespace so just the name is shown.

src/components/DefinitionEntry.js
Teju-Manchenella: If there is no revisionUrl the revision without a hyperlink. If there is no revision then nothing is shown.

src/components/DefinitionEntry.js
Teju-Manchenella: I had to do 4 to make it lineup. The InlineEditor has a pencil icon before the value to represent an editable field. In order to lineup facets with the other values, 4 spaces had to be placed.

jeffmcaffer:
Go ahead and merge this after resolving the conflicts.

PR #155, opened 2018-06-19
Remove all

Teju-Manchenella:
Remove all components from the component list.


src/reducers/listReducer.js
```diff
@@ -95,6 +95,15 @@ export default (name = '', transformer = null, comparator = null) => {
       }
     }
 
+    if (result.removeAll) {
+      return {
+        ...state,
+        sequence: ++state.sequence,
+        list: [],
+        transformedList: state.transformedList
```
jeffmcaffer: `transformedList` is essentially

```
list.map(this.transform)
```
so if you change `list`, you have to change `transformedList`

Teju-Manchenella: I am unclear how transformedList works. None of the listReducers that are created in the uiReducer pass in a transformer. My understanding of the transformedList is after the list is updated, a transformed list can be computed based on a transform function that is passed into the listReducer. Since removeAll is being used to clear all the contents of a list shouldn't the transformedList also be an empty list similar to list?

src/components/PageDefinitions.js
```diff
@@ -192,18 +197,18 @@ class PageDefinitions extends Component {
...
-          className="pull-right"
-          disabled={!this.hasChanges()}
-          onClick={this.doPromptContribute}
-        >
-          Contribute
+      <div className="pull-right">
+        <Button bsStyle="danger" disabled={!this.hasComponents()} onClick={this.onRemoveAll}>
+          Remove All
         </Button>
+        &nbsp;
```
jeffmcaffer: ok for now but should be using CSS for spacing etc rather than &nbsp;. I know it is used elsewhere. Laziness.

PR #156, opened 2018-06-19
Pr link

Teju-Manchenella:
After a successful contribution an alert pops up in the top right corner that is linked to the github pull request that was just created for the contributed curation.

### DIFF
--- a/src/actions/curationActions.js
+++ b/src/actions/curationActions.js
@@ -26,11 +26,10 @@ export function curateAction(token, spec) {
     dispatch(uiNotificationNew({ type: 'info', message: 'Started contribution.', timeout: 5000 }))
     return curate(token, spec).then(
       result => {
-        const url = `https://github.com/clearlydefined/${result.repo}/pull/${result.prNumber}`
         const prMessage = (
           <div>
             Successfully contributed{' '}
-            <a href={url} target="_blank">
+            <a href={result.url} target="_blank">
               PR#{result.prNumber}
             </a>
           </div>
--- a/src/actions/curationActions.js
+++ b/src/actions/curationActions.js
@@ -26,13 +26,7 @@ export function curateAction(token, spec) {
     dispatch(uiNotificationNew({ type: 'info', message: 'Started contribution.', timeout: 5000 }))
     return curate(token, spec).then(
       result => {
-        let url
-        switch (process.env.NODE_ENV) {
-          case 'production':
-            url = 'https://github.com/clearlydefined/curated-data/pull/' + result.prNumber
-          default:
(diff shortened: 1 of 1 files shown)


src/actions/curationActions.js
```diff
@@ -25,13 +26,31 @@ export function curateAction(token, spec) {
     dispatch(uiNotificationNew({ type: 'info', message: 'Started contribution.', timeout: 5000 }))
     return curate(token, spec).then(
       result => {
+        let url
```
jeffmcaffer: Consider replacing L29-35 with the following.

```javascript
const suffix = isDev ? '-dev' : ''
const url = `https://github.com/clearlydefined/curated-data${prefix}/pull/${result.prNumber}`
```
Not sure you can use process.env here. This code is not Node and is not running on the server. It is running in your browser. There was some funky **build time** stuff we did somewhere else (`app.js? www?`) to determine if we were in a dev environment. Should mimic that here. In fact, should likely put that in the Redux state so it just gets computed once and is available everywhere.

src/actions/curationActions.js
```diff
@@ -25,13 +26,31 @@ export function curateAction(token, spec) {
...
+        switch (process.env.NODE_ENV) {
+          case 'production':
+            url = 'https://github.com/clearlydefined/curated-data/pull/' + result.prNumber
+          default:
+            url = 'https://github.com/clearlydefined/curated-data-dev/pull/' + result.prNumber
+        }
+        const prMessage = (
+          <a href={url} target="_blank">
+            Successfully contributed PR#{result.prNumber}
```
jeffmcaffer: suggest just making the 'PR#NNN' part the link.

jeffmcaffer: suggest just making the 'PR#NNN' part the link.

src/actions/curationActions.js
Teju-Manchenella: I got the idea of using process.env.NODE_ENV from the registerServiceWorker.js and clearlyDefined.js files. In these files process.env.NODE_ENV is used to determine if the environment is development, test, or production. I tested it locally and it seems to be able to determine the correct environment (the temp variable in the code below was placed there temporarily to see what process.env.NODE_ENV is).
screen shot 2018-06-20 at 9 08 44 am

jeffmcaffer: not sure that will work in production. the code you reference in clearlyDefined.js only runs if `REACT_APP_SERVER` is not set. I believe it is. it is also true that NODE_ENV is not quite the way we expect. For example, on dev it is likely set ot production as it is a production execution environment whose purpose is to be the development environment. confusing but true.

Please validate. You'll need to check the build configurations as that env var is considered at build time.

PR #176, opened 2018-06-23
Configure testing and add some sample tests

mynameistechno:
* I researched a bit and followed `create-react-app`'s testing guide to setup something we can use to write tests going forward: 
  * https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#testing-components 
* They recommend using `enzyme` (project by airbnb) for component testing. Seems like most in the react community use this in some shape or form
* I also added a few sample tests. 
* We need to sort out what are the most important things to write tests for, to prevent regressions. E.g. we can have some shallow tests for the rendering, and spend more time on actions.
* I also added an initial travis-ci config file. @jeffmcaffer can you enable `website` in travis-ci.org? I can tweak it later.
* http://airbnb.io/enzyme/
* I also added additional enzyme matchers: https://github.com/FormidableLabs/enzyme-matchers
* And also a tool to help with mocking redux stores: https://github.com/dmitry-zaets/redux-mock-store

**Note:**
* I had a few issues getting `jest` + `enzyme` to work with the current versions of our dependecies. I ended up upgrading all of them - minor version only though! Except for `react-scripts`, which I needed the next major version, per one of the react maintainers: https://github.com/facebook/create-react-app/issues/4215#issuecomment-378330144 If needed, I can go back and upgrade only what is needed, but the site seemed fine.
* After upgrading I ran into another issue with how we were using reducers. The use of the `new` keyword in conjunction with the reducer functions was causing a `Not a constructor` errors by webpack in the dev server. It looks the reducers already return new objects using the spread operator, so I removed the `new` keyword as it didn't seem necessary, but @jeffmcaffer you may want to take a look. See commit https://github.com/clearlydefined/website/commit/3bbafb2008a199b8074e06721747909840e72840


jeffmcaffer:
@mynameistechno this is awesome! Thanks. The changes look good all up. I'll comment on your comments here.

* The version updates are good. The only thing to watch out for is react-bootstrap-typeahead. >=3.0.0 will break the harvest page. We are using an undocumented method that has been removed. See https://github.com/ericgio/react-bootstrap-typeahead/issues/266#issuecomment-378824987 if you care for the details. We can change our UI there but for right now lets stick to 2.6.0.
* the use of `new` actually is sorta weird. We should not be doing that anyway AFAICT. Curious that we are. Perhaps at one point that was the requirement and then the code never got changed. Good catch.
* as best I can tell I have enabled travis on the website repo. It seemed like it already was so ...

(BTW, somewhat unrelated but we now have contribution PR validation checks in the ClearlyDefined service so I'm going to turn off the Travis builds for PRs in the curated repos. They never got updated and are failing now. 

I'll poke at this PR locally and then merge.

PR #200, opened 2018-06-25
Sort and Filter

Teju-Manchenella:
Added sorting and filtering functionality.
Sort based on:
- License


src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
+  onSort(eventKey) {
+    this.setState({ ...this.state, activeSort: eventKey.value, counter: this.state.counter + 1 })
+    this.props.dispatch(uiBrowseUpdateList({ sort: this.getSort(eventKey) }))
+  }
+
+  filterList(list) {
+    const { activeFilters } = this.state
+    if (activeFilters.length === 0) {
+      return list
+    }
```
jeffmcaffer: excess {}

src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
+    return filter(list, component => {
+      const defintion = this.getDefinition(component)
+      for (let filterType in activeFilters) {
+        const value = activeFilters[filterType]
+        const fieldValue = this.getValue(defintion, filterType)
+        if (value === 'presence') {
+          if (!fieldValue) return false
```
jeffmcaffer: `return fieldValue`

and the `else` can be removed

src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
+        } else if (value === 'absence') {
+          if (fieldValue) return false
```
jeffmcaffer: `return !fieldValue`

src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
+        } else {
```
jeffmcaffer: `return fieldValue && fieldValue.toLowerCase().includes(value.toLowerCase())`

getting rid of negatives and nested else if etc makes the code easier to understand.

src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
+            return false
+          }
+        }
+      }
+      return true
+    })
+  }
+
+  onFilter(value) {
+    let activeFilters = Object.assign({}, this.state.activeFilters)
```
jeffmcaffer: const

src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
+    const filterValue = get(activeFilters, value.type)
+    if (filterValue && activeFilters[value.type] === value.value) delete activeFilters[value.type]
+    else activeFilters[value.type] = value.value
+    this.setState({ ...this.state, activeFilters })
+  }
+
+  incrementSequence() {
+    this.setState({ ...this.state, counter: this.state.counter + 1 })
```
jeffmcaffer: see prior naming comment

src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
   }
 
   noRowsRenderer() {
     return <div>Select components from the list above ...</div>
   }
 
-  renderContributeButton() {
+  checkSort(sortType) {
+    const { activeSort } = this.state
+    if (activeSort === sortType.value) return true
```
jeffmcaffer: `return this.state.activeSort === sortType.value`

src/components/PageDefinitions.js
```diff
@@ -181,62 +219,197 @@ class PageDefinitions extends Component {
...
+    return false
+  }
+
+  checkFilter(filterType, id) {
+    const { activeFilters } = this.state
+    for (let filterIdx in activeFilters) {
```
jeffmcaffer: consider using `some`. here it seems like you want to know if the filter is active so any entry would make it happy.

Note also, avoid `==` in favor of `===`

Also, is `filterIdx` a number?  id? ... The name is a bit off. Feel free to spell out the whole word as well.

src/reducers/listReducer.js
```diff
@@ -107,6 +111,15 @@ export default (name = '', transformer = null, comparator = null) => {
       }
     }
 
+    if (result.sort) {
+      return {
+        ...state,
+        sequence: ++state.sequence,
+        list: sort(state.list, result.sort),
```
jeffmcaffer: This should be manipulating the `transformedList`. The idea is that there is the `list` (the actual data) and the list that we use to display, the `transformedList` Right now the transformer transforms just one entry at a time rather than the whole list. If we are going to augment the reducer to do sorting, it we likely should add a `listTransformer` or some such that manages the ordering/filtering of the whole list. Then the UI just gets and displays the transformedList.

src/components/ComponentList.js
```diff
@@ -164,13 +166,24 @@ export default class ComponentList extends React.Component {
...
 
   render() {
-    const { loadMoreRows, listHeight, noRowsRenderer, list, fetchingRenderer, renderFilterBar } = this.props
+    const {
+      loadMoreRows,
+      listHeight,
+      noRowsRenderer,
+      list,
+      isFetching,
+      headers,
```
jeffmcaffer: should raise the level of abstraction here. Rather than passing in `headers`, pass in the one value we need from the headers, `hasMore`.

src/components/ComponentList.js
```diff
@@ -164,13 +166,24 @@ export default class ComponentList extends React.Component {
   }
 
   render() {
-    const { loadMoreRows, listHeight, noRowsRenderer, list, fetchingRenderer, renderFilterBar } = this.props
+    const {
+      loadMoreRows,
+      listHeight,
+      noRowsRenderer,
+      list,
+      isFetching,
```
jeffmcaffer: Looking at this now we could likely generalize a little. Rather than passing in a `noRowsRender`, a `fetchingRenderer`, and an `isFetching` flag, we can pass in a well-configured `noRowsRenderer`. We should likely make a new component (e.g., EmptyListRenderer or some such) that looks a lot like the `wrappedNoRowsRender` in `RowEntityList`. For example,

```javascript
  static defaultProps = {
    fetchingRenderer: () => <div>Loading...</div>,
    noRowsRenderer: () => <div>Nothing found</div>
  }
  render() {
    const { list } = this.props
    const renderer = list.isFetching ? fetchingRenderer : noRowsRenderer
    return <div className={'list-noRows'}>{renderer()}</div>
  }
```

This would be configured and passed in in the `render` of `PageDefinitions`

src/components/PageDefinitions.js
```diff
@@ -219,109 +227,98 @@ class PageDefinitions extends Component {
...
-  type = coordinates => {
+  type(coordinates) {
     return coordinates.type ? coordinates.type : null
   }
 
-  releaseDate = coordinates => {
+  releaseDate(coordinates) {
     const definition = this.props.definitions.entries[EntitySpec.fromCoordinates(coordinates).toPath()]
     const described = get(definition, 'described')
     if (described && described.releaseDate) return described.releaseDate
```
jeffmcaffer: `return get(definition, 'described.releaseDate')`

The one quirk is you have make sure that both `null` and `undefined` are reasonable return results

See also the `license` function

src/reducers/listReducer.js
```diff
@@ -38,6 +38,21 @@ const sort = (list, sortValue) => {
   return list ? _.sortBy(list, sortValue) : list
 }
 
+const transform = (list, transforms) => {
+  let newList = list
+  for (let transform in transforms) {
+    const transformFunction = transforms[transform].func
+    if (transform === 'sort') {
```
jeffmcaffer: should be able to leave all this to the caller. That is, it is their job to coordinate the sorting and the filtering if any. There could be arbitrary many different transformations going on. The essence of this code moves to PageDefinition where we coordinate the filter and sort settings changes to set the `transformer` on the reducer.

Note that `transformer` would be a function that takes the list and produces a list. It can do arbitrary computation to get the final result.

src/reducers/listReducer.js
```diff
@@ -82,41 +97,43 @@ export default (name = '', transformer = null, comparator = null) => {
...
-          ? update(state.transformedList, transformer(result.update), newTransformed)
-          : state.transformedList
+        list: newList,
+        transformedList: transformer ? transform(newList, transformer) : newList
       }
     }
 
-    if (result.sort) {
+    if (result.transform) {
+      transformer = _.assign(transformer, result.transform)
```
jeffmcaffer: should not need to use `assign`. just `transformer = result.transformer`

src/components/ComponentList.js
Teju-Manchenella: I am not understanding this and would like more explanation when you have time.

src/components/FilterSelect.js
```diff
@@ -0,0 +1,53 @@
...
+  { value: 'described.sourceLocation', label: 'Source' },
+  { value: 'described.releaseDate', label: 'Release Date' }
+]
+
+export default class FilterSelect extends Select {
+  static propTypes = {
+    onChange: PropTypes.func,
+    defaultFilters: PropTypes.array,
+    placeholder: PropTypes.string.isRequired,
+    disabled: PropTypes.bool
```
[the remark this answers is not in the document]
Teju-Manchenella: This component is no longer in use

PR #201, opened 2018-06-26
Collapse all

Teju-Manchenella:
Added collapse all button that collapses all the expanded components.

### DIFF
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -26,7 +26,6 @@ class PageDefinitions extends Component {
     this.onInspect = this.onInspect.bind(this)
     this.onCurate = this.onCurate.bind(this)
     this.onRemoveComponent = this.onRemoveComponent.bind(this)
-    this.doSort = this.doSort.bind(this)
     this.onChangeComponent = this.onChangeComponent.bind(this)
     this.facetChange = this.facetChange.bind(this)
     this.doPromptContribute = this.doPromptContribute.bind(this)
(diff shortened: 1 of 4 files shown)


src/components/PageDefinitions.js
```diff
@@ -186,31 +187,45 @@ class PageDefinitions extends Component {
     this.setState({ ...this.state, activeFacets })
   }
 
+  incrementSequence() {
```
jeffmcaffer: sequence or counter? Seem like this mixes naming . What is `counter`?

src/components/PageDefinitions.js
```diff
@@ -186,31 +187,45 @@ class PageDefinitions extends Component {
...
+    this.setState({ ...this.state, counter: this.state.counter + 1 })
+  }
+
   noRowsRenderer() {
     return <div>Select components from the list above ...</div>
   }
 
-  renderContributeButton() {
+  collapseComponent(component) {
+    this.onChangeComponent(component, { ...component, expanded: false })
```
jeffmcaffer: `onChangeComponent` feels like it might be heavyweight here. Not sure but the folks passing in that function may (or may not) expect it to be called for expand/collapse as the component itself is not changing.

src/components/PageDefinitions.js
```diff
@@ -237,6 +252,7 @@ class PageDefinitions extends Component {
                 githubToken={token}
                 noRowsRenderer={this.noRowsRenderer}
                 activeFacets={activeFacets}
+                counter={counter}
```
jeffmcaffer: if this is to indicate change then use `sequence`.

Issue #128, opened 2018-05-08
Implement UI support for RubyGems


fossygirl:
Once https://github.com/clearlydefined/crawler/issues/97 and https://github.com/clearlydefined/service/issues/131 have been implemented, the final piece is the UI that calls the service to finish the implementation of supporting the protocol.

Teju-Manchenella:
closed PR #153

PR #206, opened 2018-07-09
Ruby gem support

Teju-Manchenella:
add ruby gem suport


fossygirl:
Closes issue #128

src/components/RubyGemsVersionPicker.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  componentDidMount() {
+    this.getOptions('')
+  }
+
+  async getOptions(value) {
+    try {
+      const { namespace, name } = this.props.request
+      const path = namespace ? `${namespace}/${name}` : name
```
jeffmcaffer: is there every a namespace?

Teju-Manchenella: There is no namespace so I have removed any rubygem namespace references in both the service and website

PR #207, opened 2018-07-09
added support to search for packages and versions

Teju-Manchenella:
added support to search for packages and versions. added svg for nuget logo :)


src/components/NugetVersionPicker.js
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+import React, { Component } from 'react'
```
jeffmcaffer: Seems like we have this code in multiple places now. Can we generalize?

src/images/nuget.svg
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jeffmcaffer: any sense of the license or copyright for this file?

PR #210, opened 2018-07-13
renamed NuGetVersionPicker and removed namespace

Teju-Manchenella:


jeffmcaffer:
@Teju-Manchenella not sure why this build is failing. The error messages do not seem to help much. Can you check it out?

PR #211, opened 2018-07-16
fix various warnings and errors

jeffmcaffer:
With all the refactoring going on there were a number of compile warnings and runtime errors lingering. This fixes most. Some notes:

* the sort and filter button style should not be an empty string. Causes a runtime error and renders oddly in different environments. I set it to default though that is not the look we want either.
* there are still a number of `componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead,` We should figure out the right way to do this. We could just use the UNSAFE function but there might be a more generic/modern approach we should use.

### DIFF
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -3,7 +3,7 @@
 
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
-import { Grid, Row, Col, Button, DropdownButton, MenuItem, Navbar, Nav, NavItem, NavDropdown } from 'react-bootstrap'
+import { Grid, Row, Col, Button, DropdownButton, MenuItem } from 'react-bootstrap'
 import { ROUTE_DEFINITIONS, ROUTE_INSPECT, ROUTE_CURATE } from '../utils/routingConstants'
 import { getDefinitionsAction } from '../actions/definitionActions'
 import { curateAction } from '../actions/curationActions'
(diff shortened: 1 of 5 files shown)


src/components/PageDefinitions.js
```diff
@@ -346,15 +342,15 @@ class PageDefinitions extends Component {
     return (
       <DropdownButton
         className="list-button"
-        bsStyle={''}
+        bsStyle="default"
```
Teju-Manchenella: I had written 'bsStyle={''}' specifically because 'bsStyle="default"' adds unwanted styling we don't want.

bsStyle="default"
screen shot 2018-07-16 at 10 10 19 am
default style adds a box around each filter type, which is not what we want.

bsStyle=''
screen shot 2018-07-16 at 10 10 57 am
Having empty strings for bsStyle removes the box.

src/components/PageDefinitions.js
```diff
@@ -368,15 +364,15 @@ class PageDefinitions extends Component {
...
```
Teju-Manchenella: Similar comment to above

src/components/PageDefinitions.js
jeffmcaffer: right but putting in the empty string causes an error in the console. We'll need to find a different way to style BS buttons not as buttons.

Issue #84, opened 2018-03-02
Add Discord forwarder


fossygirl:
Make a https://clearlydefined.io/discord forwarder that goes to https://discord.gg/wEzHJku

PR #218, opened 2018-07-16
Adding a dicord redirect. Fixes https://github.com/clearlydefined/web…

gauntface:
Fixes https://github.com/clearlydefined/website/issues/84 

My React skills is very weak, so please check for naive mistakes here.

### DIFF
--- a/src/components/PageDiscord.js
+++ b/src/components/PageDiscord.js
@@ -1,0 +1,20 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import { ROUTE_DISCORD } from '../utils/routingConstants'
+import { uiNavigation } from '../actions/ui'
+
+class PageDiscord extends Component {
+  componentDidMount() {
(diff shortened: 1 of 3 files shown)


src/components/RehydrationProvider.js
```diff
@@ -54,6 +56,7 @@ export default class RehydrationDelayedProvider extends Component {
               <Route path={ROUTE_INSPECT} component={withTracker(PageInspect)} />
               {/* <Route path={ROUTE_CURATE} component={withTracker(PageCurate)} /> */}
               <Route path={ROUTE_HARVEST} component={withTracker(PageHarvest)} />
+              <Route path={ROUTE_DISCORD} component={withTracker(PageDiscord)} />
```
jeffmcaffer: Should be able to inline the component here and do something like 

```javascript
<Route path={ROUTE_DISCORD} component={() => window.location = 'https://discord.gg/wEzHJku'} />
```

src/components/PageDiscord.js
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
```
jeffmcaffer: Should be able to get rid of this completely (see comment re routing in RehydrationProvider)

Issue #190, opened 2018-06-25
encode current list as URL if small enough


jeffmcaffer:
Rather than forcing users to always export their list and exchange files with people, allow them to instead, get a URL that has the list (and filter/sort) encoded in it. As long as the resultant URL is not too long, it can be directly shared

jeffmcaffer:
merged

PR #219, opened 2018-07-17
View pr

daniellandau:
A view that is reachable only via a direct link from Github. Aims to show a diff in the full definition, not the diff in the curation.


daniellandau:
layout


So things marked :point_up: red can go, and show changed fields as green (as in the current PR code) and show original values in a tooltip?

Should the magnifying glass take them to `/inspect` with the PR applied, or re-enable `/curation` and show them side by side?

jeffmcaffer:
yes, @daniellandau those mods look right.

I'd suggest that changed fields in this view look the same as changed fields when they were originally changed. I believe today they have medium blue background. Should just share the style for consistency.

Yes, the magnifying glass should just go to an Inspect page with the PR. Note that the current Inspect page is bogus. It should become the Definition Details (or just Details) page. I started to mess around with something there in https://github.com/clearlydefined/website/tree/jm/detailsPage. Very sketchy but you'll start to see the idea. Again, that details page will be the same as the one users see when they are just browsing and it will allow editing of finer details (e.g., per file copyrights). As such, it should also allow people to see the full details of the proposed changes.

daniellandau:
@jeffmcaffer this PR is ready to review again, as is the other one as well.

daniellandau:
This has now code also for https://github.com/clearlydefined/website/issues/190.

src/components/DefinitionEntry.js
```diff
@@ -67,6 +67,15 @@ export default class DefinitionEntry extends React.Component {
     return get(this.props.definition, field)
   }
 
+  ifDifferent(field, then_, else_) {
+    return this.props.otherDefinition && !isEqual(get(this.props.otherDefinition, field), this.getOriginalValue(field))
```
jeffmcaffer: hmm, this pattern obscures things a bit. Perhaps just `isDifferent` and then the caller can write their own conditional. For example, in `renderWithToolTipIfDifferent` below all the rendering etc can be eliminated in the `else` case. so it would be something like 

```javascript
  if (!this.isDifferent(field)) return content
  return ( < render lots of stuff >)
```
Seems clearer and more efficient.

src/components/DefinitionEntry.js
```diff
@@ -261,71 +291,105 @@ export default class DefinitionEntry extends React.Component {
...
+              )}
             </Col>
           </Row>
           <Row>
             <Col md={2}>{this.renderLabel('Attribution', true)}</Col>
-            <Col md={10}>
-              <p className="list-singleLine">{get(licensed, 'attribution.parties', []).join(', ')}</p>
+            <Col md={10} className="definition__line">
+              {this.renderWithToolTipIfDifferent(
+                'attribution.parties',
```
jeffmcaffer: should this be `licensed.attribution.parties`? if not, why not? all the others match the prop name used in the next line for `classIfDifferent`.

src/components/DefinitionEntry.js
```diff
@@ -261,71 +291,105 @@ export default class DefinitionEntry extends React.Component {
...
             </Col>
           </Row>
         </Col>
         <Col md={7}>
           <Row>
             <Col md={2}>{this.renderLabel('Discovered')}</Col>
-            <Col md={10}>
-              <p className="list-singleLine">{get(licensed, 'discovered.expressions', []).join(', ')}</p>
+            <Col md={10} className="definition__line">
+              {this.renderWithToolTipIfDifferent(
```
jeffmcaffer: lots of boilerplate/duplicate code here. Can we make a function or something to simplify here and all the other simple entries?

src/components/PageViewPr.js
```diff
@@ -0,0 +1,78 @@
...
+import { uiNavigation } from '../actions/ui'
+import AbstractPageDefinitions from './AbstractPageDefinitions'
+
+class PageViewPr extends AbstractPageDefinitions {
+  constructor(props) {
+    super(props)
+  }
+
+  componentDidMount() {
+    const { dispatch, pr_number, token } = this.props
```
jeffmcaffer: camelCase here and everywhere

src/components/PageViewPr.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  noRowsRenderer() {
+    return <div className="placeholder-message">Fetching details on the components included in the pull request.</div>
+  }
+
+  tableTitle() {
+    const { pr_number } = this.props
+    const linkBack = this.props.base_url.isFetched ? (
+      <a href={`${this.props.base_url.item.url}/pull/${pr_number}`}>#{pr_number}</a>
```
jeffmcaffer: challenging formating. Perhaps move the `?` to a new line like

```javascript
const linkBack = this.props.base_url.isFetched 
  ? (<a href={`${this.props.base_url.item.url}/pull/${pr_number}`}>#{pr_number}</a> ) 
  : (`#${pr_number}`)
```

src/components/PageViewPr.js
```diff
@@ -0,0 +1,78 @@
...
+
+import React from 'react'
+import { connect } from 'react-redux'
+import { Button } from 'react-bootstrap'
+import { uiViewPrGetData, uiViewPrGetBaseUrl, uiViewPrUpdateList } from '../actions/ui'
+import { ROUTE_VIEW_PR } from '../utils/routingConstants'
+import { uiNavigation } from '../actions/ui'
+import AbstractPageDefinitions from './AbstractPageDefinitions'
+
+class PageViewPr extends AbstractPageDefinitions {
```
jeffmcaffer: This ought to be `PageContribution` and the URL should be `clearlydefined.io/contributions/174`

"view pr" is too implementation specific.

This naming change will affect a lot of naming. like the action names, the route name and value, ...

src/components/PageDefinitions.js
```diff
@@ -1,100 +1,118 @@
...
+    this.doSaveAsUrl = this.doSaveAsUrl.bind(this)
   }
 
-  getDefinition(component) {
-    return this.props.definitions.entries[EntitySpec.fromCoordinates(component).toPath()]
+  componentDidMount() {
+    const { dispatch, path } = this.props
+    if (path.length > 1) {
+      try {
+        const str = pako.inflate(base64js.toByteArray(path), { to: 'string' })
```
jeffmcaffer: prefer full, semantic naming. e.g `definitionSpec`

src/components/PageDefinitions.js
```diff
@@ -1,100 +1,118 @@
...
+  }
+
+  doSaveAsUrl() {
+    const { dispatch, components } = this.props
+    const spec = this.buildSaveSpec(components.list)
+    const fileObject = { filter: this.state.activeFilters, sortBy: this.state.activeSort, coordinates: spec }
+    const url = `${document.location.origin}/definitions/${base64js.fromByteArray(
+      pako.deflate(JSON.stringify(fileObject))
+    )}`
+    const textArea = document.createElement('textarea')
```
jeffmcaffer: consider factoring this out as a "copyToClipboard" function

daniellandau:
Thank you for the thorough review. Both pull requests now updated according to comments, I hope I didn't miss anything.

src/actions/prActions.js
```diff
@@ -0,0 +1,50 @@
...
+              current,
+              path: component,
+              component: EntitySpec.fromPath(`${component}/pr/${pr_number}`)
+            }))
+          )
+        )
+      )
+      .then(result => {
+        dispatch(actions.success(result))
+        result.forEach(x => dispatch(uiViewPrUpdateList({ add: x.component })))
```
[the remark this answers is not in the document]
Teju-Manchenella: Hey @daniellandau sorry for the late response, my github notifications are not working properly. I am not sure how you have gone ahead with bulk adding components; however, I have a PR #232 that addresses bulk adding components.

jeffmcaffer:
Thanks @daniellandau . this is great! There is some inconsistency in the naming of contributions, curations, ... but that is not your fault, the current code is somewhat messed up. I will go through and do some fixes.

PR #220, opened 2018-07-18
Validation message

Teju-Manchenella:
Edited the handleResponse() function in clearlyDefined.js. So far when the website gets an error it simply receives an error status. Contribution validation; however, sends over both an error status and an error body. Previously the handleResponse() function was not resolving error bodies. Now the handleResponse() function has the ability to resolve error bodies so the can be used.

Each contribution error appears as a UI message notification now as seen in the picture below. The error messages are based on the new error handling that was introduced in service PR [169](https://github.com/clearlydefined/service/pull/169). This is my first attempt at notifying the user that their contribution is incorrect. Any suggestions will be helpful.

screen shot 2018-07-18 at 3 59 56 pm

### DIFF
--- a/src/actions/curationActions.js
+++ b/src/actions/curationActions.js
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 import React from 'react'
+import { flatten } from 'lodash'
 import { asyncActions } from './'
 import { curate, getCuration } from '../api/clearlyDefined'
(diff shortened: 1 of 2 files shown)


jeffmcaffer:
Looks good. eventually we will put the errors right in place in the UI rather than in notification boxes and in fact, the validation should be happening before the call even happens so this is really a backstop scenario. This approach is just fine.

Issue #191, opened 2018-06-25
File list view


jeffmcaffer:
The full details (and other) views should show the list of files that are part of the given definition.  All the per-file information such as the following should be shown

* The path of the file
* The list of facets in which the file appears
* The list of licenses discovered in the file
* The list of copyright statements found in the file

The list should allow relatively easy/natural navigation as it may have many entries. The current thinking is a tree-table style where the left column is a directory tree and the subsequent columns are the above data per file.

That the file list allows sorting and filtering much like the component list is super useful and coherent. Seeing the list of files like I would in a normal operating system file tree view (with expand and collapse) makes me feel at home.  (Note: TBD how the tree view changes in the face of sorting. Filtering could just make it more sparse but sorting changes the structure)

fossygirl:
We have some mockups for the evolution of this list view. 

The first is a read-only view:

![read only tree table view copy with hover](https://user-images.githubusercontent.com/17101092/42352413-37f24336-806f-11e8-9d0a-fb8e52626000.png)

Second phase is a read/write view. Facets should always be a dropdown since there are only 6 and those won't change. Licenses will be expressions, so when we get to this phase we will need a [license](https://github.com/clearlydefined/website/issues/178) and [facet picker](https://github.com/clearlydefined/website/issues/202) as per their respective issues. 

![read_write tree table view](https://user-images.githubusercontent.com/17101092/42352419-3ffd690c-806f-11e8-85fc-fe6608fd277c.png)

Third phase is a filtering option to let users filter on facets or licenses:

![filter on columns](https://user-images.githubusercontent.com/17101092/42352440-5db18938-806f-11e8-8972-f7312217cf07.png)

storrisi:
I've built a sample on stackblitz to test if react-table could do the trick.
The first implementations had some problems, due to the way that react-table groups data in a pivot column.
The problems occur when some folded data has some children different from branch to branch.
react-table assumes that every branch should have the same number of children, so if one them doesn't, it renders an undefined.

I've spent some hours to find how I could overcome this behaviour, and the way I have found seems to fix the problem.

I've started from the HOC TreeTable component that react-table team released, and I've built a customized copy in my project.

Next, I've analyzed the structure of the object passed to the Renderer, and written a function that removes any undefined unnecessary object, 
re-structuring the proper tree-object that should be rendered.

Here you can find a sample of what I've implemented:
https://stackblitz.com/edit/react-ihqxyx

Some other problems need to be solved to accomplish the task.

[...]

PR #223, opened 2018-07-21
FileList View Component

storrisi:
First solution provided related to #191


storrisi:
This PR implements a solution to #191 
It uses `react-table` to show a TreeView with Table Columns. 
Each column is filterable, and extended column details are shown by the use of a Popover.

src/components/FileList/FacetsRenderer.js
```diff
@@ -0,0 +1,16 @@
...
+export default class FacetsRenderer extends Component {
+  render() {
+    const { item } = this.props;
+    return (
+      <div>
+        {item.value}
+      </div>
+    )
+  }
+}
```
mynameistechno: @jeffmcaffer thoughts on using the functional style of components when they are stateless? It's more concise. I'd write this as 
```js
const FacetsRenderer = ({item}) => <div>{item.value}</div>;
```
Internally we've been using `flow` for prop validation, but with prop types I think you'd do this
```js
FacetsRenderer.propTypes = { item: PropTypes.object };
export default FacetsRenderer;
```

src/components/FileList/FileList.js
```diff
@@ -0,0 +1,155 @@
...
+    const { files, isFiltering } = this.state;
+
+    return <div>
+      <TreeTable
+        showPagination={false}
+        sortable={false}
+        filterable={true}
+        freezeWhenExpanded={false}
+        manual={false}
+        onFilteredChange={(filtered) => this.setState({ isFiltering: true })}
```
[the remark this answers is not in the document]
mynameistechno: prettier should take care of this I think

src/components/FileList/FileList.js
```diff
@@ -0,0 +1,155 @@
...
+
+// Import Custom TreeTable HOC
+const TreeTable = treeTableHOC(ReactTable);
+
+const pathColums = [];
+const columns = [];
+
+// Parse Path to retrieve the complete folder structure
+const parsePaths = (data) => {
+  const res = data.map(item => {
```
mynameistechno: Can you just return here? `return data.map(item => {`

src/components/FileList/FileList.js
```diff
@@ -0,0 +1,155 @@
...
+        }}
+      />
+    </div>
+  }
+}
+
+// Import Custom TreeTable HOC
+const TreeTable = treeTableHOC(ReactTable);
+
+const pathColums = [];
```
mynameistechno: typo? `pathColumns`

src/components/FileList/FileList.js
```diff
@@ -0,0 +1,155 @@
...
+    } else {
+      folders.unshift("/")
+    }
+
+    //Add item[`folder_${index}`] to item object
+    //If index is the last item, then is the name of the file
+    folders.forEach((p, index) => { if (index + 1 === folders.length) item.name = p; else { item[`folder_${index}`] = p; } })
+
+    folders.forEach((p, index) => {
+      if (index + 1 < folders.length && pathColums.indexOf(`folder_${index}`) === -1) {
```
mynameistechno: If you don't need the index maybe 
```
!pathColums.includes(`folder_${index}`)
```
is more readable?

src/components/FileList/FacetsRenderer.js
jeffmcaffer: I do like the conciseness of the functional style components. OTOH I like the regularity of having a class. I'm flexible here but would like there to be some logic to the choice. Eg. if small and stateless, functional, else class. (whatever)

src/components/FileList/FilterCustomComponent/FilterCustomComponent.js
```diff
@@ -0,0 +1,38 @@
...
+   * 
+   */
+  changeFilterValue = (filterValue) => {
+    this.setState({ filterValue: filterValue.target.value }, () => this.props.onChange(this.state));
+  }
+
+  render() {
+    return (
+      <FormGroup className="inputBox">
+        <FormControl type="text" onChange={this.changeFilterValue} value={this.props.filter ? this.props.filter.value.filterValue : ''} />
```
mynameistechno: For checking nested objects I like to use a helper like lodash's `get`, e.g. `_.get(object, 'a.b.c', 'default')`
`_.get(this.props, 'filter.value.filterValue', '');`

src/components/FileList/treeTable.js
```diff
@@ -0,0 +1,104 @@
...
+     * This is the custom implementation where each subRow with _pivotVal "undefined" is parsed
+     * in any of his _subRows childs, until find the first _pivotVal !== "undefined"
+     * 
+     * Once the object has been found, than it will returned to the first parent, 
+     * updating the value of _nestingLevel as well
+     * 
+     */
+    getSubrows = subRows => {
+      subRows && subRows.forEach((subRow, index) => {
+        if (subRow._pivotVal === "undefined" && subRow._subRows) {
```
mynameistechno: So the value of the string will be "undefined"? or might the value be `undefined`? Just making sure! 😜

mynameistechno:
I took a quick look over the code and left some minor comments. I can't say I looked very closely at the logic, there's a lot going on here. Some smaller functions and unit tests would help reason about the code for anybody that needs to jump in. However I should note the codebase as a whole is lacking in the tests department 😝- we're working on it though! Having said that, I played around with the demo at https://react-uji3fp.stackblitz.io/ and it seems to work pretty well!

src/components/FileList/treeTable.js
storrisi: yes, "undefined" is the value of the string. This is a fix i implemented to solve a wrong behaviour of react-table that generates some "undefined" folders

src/components/FileList/CopyrightsRenderer/CopyrightsRenderer.css
```diff
@@ -0,0 +1,3 @@
+.copyrightPopover{
```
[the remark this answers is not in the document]
storrisi: I preferred to keep the style attached to the component so it could be easily found and managed. Not a big deal to move on another place for consistency

storrisi:
PR is updated solving the fixes reported during the Code Review

PR #232, opened 2018-07-25
add all components in single UI render

Teju-Manchenella:
UI only rendered once when components loaded from file. Adding this along with moving the badge logic to frontend makes the site much more responsive and can quickly process large component lists (ex. package-lock.json files).


jeffmcaffer:
likely obsoleted by #234 due to refactoring under the hood

Teju-Manchenella:
Changes added to pr #234. Closing request.

PR #234, opened 2018-07-31
port addAll changes to new defintions page

jeffmcaffer:
@Teju-Manchenella did some changes in #232 but the code has been refactored underneath those changes. This ports the changes to make dropping a list of coordinates happen as a batch rather than one at a time.

@daniellandau you might want to take a look as well.

### DIFF
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -95,7 +95,7 @@ class PageDefinitions extends AbstractPageDefinitions {
   }
 
   doSaveAsUrl() {
-    const { dispatch, components } = this.props
+    const { components } = this.props
     const spec = this.buildSaveSpec(components.list)
     const fileObject = { filter: this.state.activeFilters, sortBy: this.state.activeSort, coordinates: spec }
     const url = `${document.location.origin}/definitions/${base64js.fromByteArray(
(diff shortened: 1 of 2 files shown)


src/components/PageDefinitions.js
```diff
@@ -176,40 +176,25 @@ class PageDefinitions extends AbstractPageDefinitions {
...
+    dispatch(uiBrowseUpdateList({ addAll: toAdd }))
+    const missingDefinitions = toAdd.map(spec => spec.toPath()).filter(path => !definitions.entries[path])
+    await dispatch(getDefinitionsAction(token, missingDefinitions))
+    dispatch(
+      uiBrowseUpdateList({
+        transform: this.createTransform.call(
+          this,
+          listSpec.sortBy || this.state.activeSort,
+          listSpec.filter || this.state.activeFilters
+        )
```
Teju-Manchenella: This is an interesting case if the user has selected a set of sorts/filters and then drops a file that has its own set of sorts/filters. I think it makes more sense to use the sorts/filters that are in the file as opposed to the sorts/filters that the user selected. I think most users probably won't have any components or sorts/filters selected before dropping a file. I think naturally a user would go to ClearlyDefined.io and if they had a file to drop they would drop it immediately before adding any other components or sorts/filters. There is no real solution to this but just something we can keep in mind.

Issue #231, opened 2018-07-25
Details view


jeffmcaffer:
Need a reusable component that shows the full detail of a definition. It should optionally allow editing of the values in definition in a way that is shareable with other components. That is, the detail component described here does not "own" the definition. This component will show up in two primary places/ways

1. as a popup/overlay on the main Definitions page when the user wants to dive into a given definition in the list
1. as an independent page with a deep link URL allowing users to talk about definitions and reference them easily

A mockup of the page is in the jm/detailsPage branch. This is just a **mockup** and is not intended to be the implementation. There is tons of duplicate code etc

some key points
* the details component should be standalone so it can be used in page, an overlay, ... 
* Over time we will undoubtedly redesign the layout.
* As the user progresses from the two line entry to the expanded panel to the full details view of the component, the experience should be smooth. With visual and content continuity. They are just looking at more data about the same thing so the data they have already seen should be similar.

There are a few new components and related features that will show up here
* Facet picker (https://github.com/clearlydefined/website/issues/202)
* File List (https://github.com/clearlydefined/website/issues/191)
* Editing (https://github.com/clearlydefined/website/issues/196)

AlexWebYourmind:
Hi @jeffmcaffer (and all).
We have just completed an Adobe XD prototype showing the main flow (Read / Edit / Contribute) for the Full Detail Screen (Modal view, DeepLinking version will be added soon)

The prototype is available at https://xd.adobe.com/view/1fef18d7-16f1-4264-57f3-75cb6d1cf9ac-69de/

We have a few questions around the Deeplinking version, one of which being:
- Should we add a contribute button when the full details is visited directly (deeplinking) as the flow will not be started from the "Browse" Page?

@jeffmcaffer please review the prototype, @storrisi and @gianpaj will start actual implementation this week. As long as we are all happy with the protoype, we might use it as our goto document when we need to review the UX on that section.

jeffmcaffer:
@AlexWebYourmind prototype looks good. There were a few things we talked about in the team call just now but overall, +1. 

For deep linking 

* The contribute button should be on the full details PAGE but not on the OVERLAY version. 
* Separately you asked about the URL for the deep link. From a user point of view it seems to make sense to have clearlydefined.io/definitions/npm/npmjs/-/foo/1.0 as the link. I'm personally not worried about the lack of RESTfullness of the plain clearlydefined.io/definitions page. The UI is not a REST api. We might need to do something interesting in the routing or just have a switch in the `render`.

[...]

PR #241, opened 2018-08-07
Full Details View - Read Only

storrisi:
Related to issue #231


src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+
+  handleSave = changes => {
+    const { onSave, component } = this.props
+    const newComponent = { ...component, changes }
+    onSave && onSave(component, newComponent)
+  }
+
+  handleClose = () => {
+    const { onClose } = this.props
+    onClose()
```
jeffmcaffer: is `onClose` required? (no proptypes above) If not, need to protect against null/undefineds

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+import ContributePrompt from '../ContributePrompt'
+import FullDetailComponent from './FullDetailComponent'
+
+/**
+ * Component that renders the Full Detail View as a Page or as a Modal
+ * based on modalView property
+ *
+ */
+export class FullDetailPage extends Component {
+  componentDidMount() {
```
jeffmcaffer: declare proptypes

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+    const { path, uiNavigation, component } = this.props
+
+    if (path && component) {
+      this.handleNewSpec(component)
```
jeffmcaffer: unnecessary {}

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+    if (path && path !== this.props.path) {
+      this.handleNewSpec(component)
+    }
+  }
+
+  /**
+   * Get the data for the current definition
+   *
+   */
+  handleNewSpec = component => {
```
jeffmcaffer: I'm honestly torn between arrow functions and binding functions in the constructor. In the rest of the code we use bind. In general I error on the side of consistency.  Thoughts?

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+    const { token, uiInspectGetDefinition, uiInspectGetCuration, uiInspectGetHarvested } = this.props
+    if (component) {
```
jeffmcaffer: nit alert

I'd tend towards test and exit and then do the full functionality. So

```javascript
if (!component) return
const {}
....
```
Part of a general pattern of lower apparent complexity.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+  handleClose = () => {
+    const { onClose } = this.props
+    onClose()
+  }
+
+  render() {
+    const { path, component, definition, curation, harvest, modalView, visible } = this.props
+
+    return modalView ? (
+      <Modal
```
jeffmcaffer: consider breaking this out into the following just to make the logic clearer

```javascript
return modalView ? this.renderModal() : this renderStandard()
```

or some such.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+}
+
+function mapDispatchToProps(dispatch) {
+  return bindActionCreators(
+    { uiInspectGetDefinition, uiInspectGetCuration, uiInspectGetHarvested, uiNavigation, curateAction },
+    dispatch
+  )
+}
+
+FullDetailPage.propTypes = {
```
jeffmcaffer: Put these at the top as a static class prop

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+import Contribution from '../../utils/contribution'
+import ContributePrompt from '../ContributePrompt'
+import FullDetailComponent from './FullDetailComponent'
+
+/**
+ * Component that renders the Full Detail View as a Page or as a Modal
+ * based on modalView property
+ *
+ */
+export class FullDetailPage extends Component {
```
jeffmcaffer: This is not necessarily a "page" rather, it is a view where it may be embedded in a modal or just be a standard component. Seems like there is relatively little shared code here. Much of it relates to either the modal case (onClose, ...) and the other parts are only used in the standard case. 

Consider splitting into two?

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -0,0 +1,170 @@
...
+
+function mapDispatchToProps(dispatch) {
+  return bindActionCreators(
+    { uiInspectGetDefinition, uiInspectGetCuration, uiInspectGetHarvested, uiNavigation, curateAction },
+    dispatch
+  )
+}
+
+FullDetailPage.propTypes = {
+  /* Define if the visualization should be as a Modal or as a Page */
```
jeffmcaffer: convention has been to use // comments

src/utils/routingConstants.js
```diff
@@ -3,6 +3,7 @@
 
 export const ROUTE_ABOUT = '/about'
 export const ROUTE_DEFINITIONS = '/definitions'
+export const ROUTE_DETAIL = '/detail'
```
jeffmcaffer: Thought we were going to do `/definitions/npm/npmjs/-/foo/1.0` for the route here? need a bit of routing work but that keeps it simple for users. /definitions gets them all the defs, /definitions/... gets a specific one.

src/actions/curationActions.js
```diff
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: MIT
 
 import React from 'react'
-import { flatten } from 'lodash'
+import flatten from 'lodash/flatten'
```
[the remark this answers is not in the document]
storrisi: The main advantage is on the final bundle size, using this method only the required module will be really imported by webpack.
You can see here a test of performances on this specific behaviour: https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark

src/components/AbstractPageDefinitions.js
```diff
@@ -92,9 +93,22 @@ export default class AbstractPageDefinitions extends Component {
     this.props.history.push(url)
   }
 
+  /**
```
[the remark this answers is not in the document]
storrisi: Make sense, we will change the comments approach

src/components/FullDetailView/FullDetailPage.js
storrisi: The two methods are exactly the same. The arrow functions syntax is compiled to the bind method.
I always preferred the arrow functions syntax as a standard, but for consistency of the project we can switch to the bind method.

src/components/FullDetailView/FullDetailPage.js
storrisi: Originally this was splitted into 2 components.
But since the methods for connect to the Redux store are the same (for saving, contribution), i've preferred to merge into one component, to keep it easily maintainable

src/utils/routingConstants.js
storrisi: That kind of control has been done on the Router definition, on the ReydrathionProvider.

Indeed on that component we've splitted into 2 behaviour the /definitions  routes.
Using "exact" prop of react-router, we've assumed that when the route is exactly "/definitions" then it will show the Browse Page.
Otherwise, if the url contains some other data, it will show the FullDetailPage.

This make sense for you?

src/components/FileList/FileList.js
```diff
@@ -146,8 +152,22 @@ const pathColums = []
 const columns = []
 
 // Parse Path to retrieve the complete folder structure
-const parsePaths = data => {
-  return data.map(item => {
+const parsePaths = (data, changes, component) => {
+  const changedFacets = pickBy(changes, (item, index) => index.startsWith('described.facets'))
+  return data.map((item, index) => {
+    item.facets = map(changedFacets, (glob, facets) => {
```
jeffmcaffer: stepping back from the details of this particular function, when the user changes the facet definition we need to go back to the server and recompute the definition based on those changes. Facet changes influence more than just the tags on the files. The set of discovered licenses and copyrights, the number of missing files and indeed the scores can all change as a result of facet changes. There is an API you can Post to do a `preview` basically saying "here is a proposed patch, what would the definition look like". Given the changed definition then we need to compute the delta and present those values as changed.

src/components/FullDetailView/FullDetailPage.js
jeffmcaffer: The model does not have contribute and the standard view should not have save. I'm actually not sure what code would be duplicated between the two if they were separate.

src/components/FullDetailView/FullDetailPage.js
jeffmcaffer: Actually my comment here would be addressed if this were actually two different components.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -20,31 +19,49 @@ import FullDetailComponent from './FullDetailComponent'
...
+    this.doPromptContribute = this.doPromptContribute.bind(this)
+    this.handleSave = this.handleSave.bind(this)
+    this.handleClose = this.handleClose.bind(this)
+    this.onChange = this.onChange.bind(this)
+    this.getValue = this.getValue.bind(this)
+    this.classIfDifferent = this.classIfDifferent.bind(this)
+  }
+
+  static propTypes = {
+    //Define if the visualization should be as a Modal or as a Page
```
jeffmcaffer: total nit: space after the // please

src/utils/contribution.js
```diff
@@ -42,11 +44,38 @@ export default class Contribution {
     return result
   }
 
-  /**
-   * Check if the current component has listed changes
-   *
-   */
-  static hasChange = entry => {
+  // Check if the current component has listed changes
```
jeffmcaffer: commenting nit: I'd rather there function and parameter names be good than have to add comments like this. For example, `hasChange` is pretty clear that it is testing for changes. Indeed it is testing the provided argument, not the current component. So if the function were `hasChanges(component)` then it really ought not even need a comment. 

Similar feed back on the comments for the other functions.

Summary being that if we have to comment simple things like this then the naming or structuring could probably be improved.

src/components/FacetsEditor/FacetsEditor.js
```diff
@@ -0,0 +1,57 @@
...
+    definition: PropTypes.object,
+    changes: PropTypes.object
+  }
+
+  render() {
+    const { onChange, getValue, classIfDifferent } = this.props
+
+    return (
+      <div className="facetsEditor">
+        <h1>Facets</h1>
```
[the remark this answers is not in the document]
storrisi: Sure, this is a first raw implementation

src/actions/curationActions.js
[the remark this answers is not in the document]
jeffmcaffer: fair enough. For this style to be effective, do we need to do this everywhere? If so, let's be consistent until we decide to do that optimization across the entire codebase. 

Consistency wins! Otherwise the code gets really messy.

src/components/FileList/FileList.js
```diff
@@ -146,8 +152,22 @@ const pathColums = []
 const columns = []
 
 // Parse Path to retrieve the complete folder structure
-const parsePaths = data => {
-  return data.map(item => {
+const parsePaths = (data, changes, component) => {
+  const changedFacets = pickBy(changes, (item, index) => index.startsWith('described.facets'))
+  return data.map((item, index) => {
```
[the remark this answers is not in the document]
storrisi: Yes, data is the definition.files object, that will be parse by this function (it was already on the FileList component).
Given that, we could definetly rename them for consistency

src/components/FileList/FileList.js
storrisi: So, if i've understand the point, we don't need to apply the changes of a facets only at client side, but we need to post the data to this preview API and receive back the results.
It this right?
Is this behaviour appliable only to facets, or it's something that we should do for each kind of change on the definition? (as for copyrights and licenses as well)

src/utils/contribution.js
storrisi: You are totally right, we will do some improvements on this.

src/actions/curationActions.js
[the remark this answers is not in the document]
AlexWebYourmind: @jeffmcaffer
Sorry to jump in in this but I want to make sure we are all on the same page.

While I do agree that Consistency wins, I do also think that if we want to introduce improvements, best practices, etc. it could be beneficial to "lead by example" and optimize new code, while being mindful of avoiding "messy code"

We could then provide evidence of th3 benefit and address refactoring for existing code.

Given the nature of the lodash change, may i suggest to bring it in?

Otherwise, let us know your thoughts and how do you think we should approach the introduction of new best practices, refactoring etc from a process perspective. 

Thanks!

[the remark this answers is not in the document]
jeffmcaffer: What I want to avoid is that good ideas for programming style come up, get used in a bit of the code that is touched and then never ends up spreading. That leaves us with an inconsistent codebase and no tangible improvements. 

To execute on this I see two approaches. 

1. Augment the coding practices doc to detail the choices and explicitly call out that they will be incrementally introduced. Ideally with some sort of target time and commitment to get it done.
2. Do a full on sweep of the code and just change it all at once. Would be good to time those things for a time when it won't generate a lot of merge conflicts but sure.

This is especially true of changes that need to be pervasive to be effective. 

Either is fine with me but honestly at this point optimizing out a few K of lodash is way low on the priority list. Even as a general practice, until we see that load time is an issue (eg., when we go to support mobile) to me consistency still wins.

src/components/FileList/FileList.js
jeffmcaffer: Great point @storrisi . There are a set of changes that affect computed values in facets. Things like file copyrights and licenses for example. Now, some of these are easy to recompute on the client (e.g., the number of files missing licenses). It would be more consistent to hit the server and have it recompute the definition given a proposed set of changes for the definition. We'll have to look at the perf cost and likely do a bit of caching in the service. 

I've not completely worked through all the implications but would suggest that we try that approach first and see if there are any blockers. Make sense?

src/utils/routingConstants.js
jeffmcaffer: It does make sense but `ROUTE_DETAIL` is not used anywhere that I can see. It is imported in the RehydrationProvider bu not used.

src/actions/curationActions.js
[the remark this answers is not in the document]
AlexWebYourmind: @jeffmcaffer Thanks for your repy!

I couldn't agree more with all of the above and agree on the fact that the lodash change is very low priority. 


As for the process, I think we all agree. 

Extending the documentatiom definitely make sense when introducing coding pratices and styleguides, probably less when tuning a very specific bit of code.

And yes, we will make sure to be consistent, and make sure to apply improvents anywhere needed at once, as long as that doesn't cause huge merge conflicts or has a big impact on timelines etc.

storrisi:
This PR can now be reviewed

Issue #202, opened 2018-06-28
Facet picker 


fossygirl:
Just like the license and date pickers we want to use in our UIs for various views, we want to have a facet picker that lists the facets and allows users to select a facet. The main scenario for this is on the Full Details page for a component.

We will likely need to iterate on the design/flow of this. Starting off simple, perhaps something and leveraging the following assumptions:
* There are exactly 6 facets
* The `core` facet is not user-defined. Rather it is whatever is left over from not being in another facet
* Each facet is a list of globs
* Facets can overlap. that is, one file can be in multiple facets
* A very common case is that a folder and its transitive children are assigned to a facet

So a first cut might be something like 

facet | definition
----|---------
core | "everything that is not otherwise assigned a facet"
data | **/*.csv
dev |
doc | 
examples | src/examples/**/*.js
test | test/**/*.js, testrunner.js

Where
* the right side is just and editable CSV with each entry being a valid glob
* the right side may be empty. in fact it will typically start empty
* the core entry is a fixed string with content roughly as shown

Note: The table format above is just illustrative. Not intending this compoentn to show up actually as a table with borders etc. 

## Evolutions
As we go forward this component can get richer and richer. We should prioritize improvements based on user feedback.

1. Facet specs are validated
1. rather than being a plain, single line text input, the glob spec is a multi-line (one CSV element per line) UI affordance with the ability to add, delete, and modify rows. Each entry would have content assist for helping to complete the folder names etc.
1. As the user edits the entries above, the entries in the filelist are updated with their recomputed facet tags. If the file list is filtered by facet then the user will see just the files that are in the selected facet. That allows them to see the effect of their facet edits.
2. In the file list a user can select a **folder** and tag it with one or more facets. that would compute the relevant glob and add it to the specified facet(s).
1. Potentially users are able to click on a file in the file list and mark it with facets. The challenge here is in creating and maintaining the facet spec if the user is enumerating many individual files rather than creating a glob pattern

jeffmcaffer:
addressed by #248

Issue #196, opened 2018-06-25
in situ editing in detail view


jeffmcaffer:
When looking at the full detail for a definition the user should be able to do in situ editing and contribution for every possible value.

The full detail view is augmented to highlight the fields that have pending changes and all the normal field change affordances are available. 

there is a familiar (from the definition list) Contribute button that allows me to describe my changes before submitting the PR.

PR #248, opened 2018-08-16
Full Details View - Editing experience

storrisi:
Implementation for the editing of a Definition.

It includes:
- Facet picker (#202)
- File List (#191)
- Editing (#196)


src/api/clearlyDefined.js
```diff
@@ -12,7 +12,7 @@ export const apiHome = process.env.REACT_APP_SERVER || getServiceDefaultUrl()
 function getServiceDefaultUrl() {
   switch (process.env.NODE_ENV) {
     case 'development':
-      //TODO this need to be replaced when using a local enviroment, it should become API_LOCAL
+      // TODO this needs to be replaced when using a local enviroment, it should become API_LOCAL
```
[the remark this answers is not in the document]
storrisi: I've updated the file setting the instruction you mentioned, adding a comment in order to let the developer know the that file mustn't be committed if he changed the process.env value

storrisi:
The PR can new be reviewed and merged.

src/components/FileList/FileList.js
```diff
@@ -64,16 +56,20 @@ export default class FileList extends Component {
...
-              item._original && item._original.facets
-                ? item._original.facets
-                    .toString()
-                    .toLowerCase()
-                    .includes(filter.value.filterValue.toLowerCase())
-                : true
-          ),
+        filterMethod: (filter, rows) => {
+          const filterValue = filter.value.filterValue.toLowerCase()
+          return rows.filter(item => {
```
jeffmcaffer: consider something like (note `find` from lodash)
```javascript
return rows.filter(item => {
  if (!get(item, ‘facets.length’)) return true
  return find(item.facets, f => f.value.includes(filterValue))
}
```

src/components/FileList/FileList.js
```diff
@@ -152,30 +151,36 @@ const pathColums = []
 const columns = []
 
 /**
- * Parse Path to retrieve the complete folder structure
- * @param  {} files
- * @param  {} changes
- * @param  {} component
+ * Parse each file's path to retrieve the complete folder structure
+ * @param  {} files The files object coming from the definition
```
jeffmcaffer: other parameters?

src/components/FileList/FileList.js
```diff
@@ -152,30 +151,36 @@ const pathColums = []
...
-        .pop()
-        .trim()
-    })
-
-    file.areFacetsDifferent =
-      file.facets.length > 0 && isEqual(Contribution.getOriginalValue(component, `files[${index}].facets`), file.facets)
-        ? ''
-        : 'facets__isEdited'
-
+const parsePaths = (files, component, preview) => {
```
jeffmcaffer: this whole method has quite high apparent complexity. Can it be simplified and/or broken into reasonably named sub methods?

src/components/FileList/FilterCustomComponent/FilterCustomComponent.js
```diff
@@ -49,7 +35,7 @@ FilterCustomComponent.propTypes = {
    */
   filter: PropTypes.object,
   /**
-   * Callback function
+   * React-Table callback function
```
jeffmcaffer: not sure what this comment means or how it helps.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -57,71 +67,118 @@ export class FullDetailPage extends Component {
 
   componentWillReceiveProps(nextProps) {
     const { path, component } = nextProps
-    if (path && path !== this.props.path) this.handleNewSpec(component)
+
+    if (path && path !== this.props.path) {
```
jeffmcaffer: consider the more compact and less nested...

```javascript
    if (!path || path === this.props.path) return 
    if (component.changes)  return this.setState({ changes: component.changes }, () => this.handleNewSpec(component))
    this.handleNewSpec(component)
```

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -57,71 +67,118 @@ export class FullDetailPage extends Component {
...
 
   // Get the data for the current definition
   handleNewSpec(component) {
     const { token, uiInspectGetDefinition, uiInspectGetCuration, uiInspectGetHarvested } = this.props
-    if (component) {
-      uiInspectGetDefinition(token, component)
-      uiInspectGetCuration(token, component)
-      uiInspectGetHarvested(token, component)
-    }
+    if (!component) return false
```
jeffmcaffer: does a value need to be returned here? A bit misleading unless it is meant to mean something.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -57,71 +67,118 @@ export class FullDetailPage extends Component {
...
-    this.setState({ visible: false })
-  }
-
-  // Function called when a data has been changed
-  onChange(item, value) {
-    const { component } = this.props
     const { changes } = this.state
-    this.setState({ changes: Contribution.onChange(component, changes, item, value) }, () =>
-      console.log(this.state.changes)
+    if (isEmpty(changes)) return onClose()
```
jeffmcaffer: Break this out into a reuasable chunk of code (component or otherwise) then just call if from here.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -57,71 +67,118 @@ export class FullDetailPage extends Component {
...
+    const key = `open${Date.now()}`
+    const btn = (
```
jeffmcaffer: full names (e.g., `button` please. 

On a separate note, seems complicated for a button. Perhaps the name could be even better?

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -57,71 +67,118 @@ export class FullDetailPage extends Component {
...
+  close() {
+    const { uiCurateResetDefinitionPreview, onClose } = this.props
+    this.setState({ changes: {} }, () => {
+      uiCurateResetDefinitionPreview()
+      onClose()
+    })
   }
 
-  classIfDifferent(field, className) {
+  // Function called when a data has been changed
```
jeffmcaffer: `onChange` means exactly what that comment says so the comment is redundant. I'm all for comments where they add value. otherwise they are just some more code that needs to be maintained. So many times I've been led astray by "helpful" comments that in the end were wrong because the code changed and the comment didn't.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -130,38 +187,39 @@ export class FullDetailPage extends Component {
         centered
         destroyOnClose={true}
         visible={visible}
-        onOk={this.handleSave}
-        onCancel={this.handleClose}
         width={'85%'}
```
jeffmcaffer: can we also add some top and bottom margin? Without it feels a bit odd having it fill the vertical space as well.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -57,71 +67,118 @@ export class FullDetailPage extends Component {
...
+      this.previewDefinition()
+    )
   }
 
   render() {
-    const { path, component, definition, curation, harvest, modalView, visible } = this.props
+    const { path, definition, curation, harvest, modalView, visible, previewDefinition } = this.props
     const { changes } = this.state
-
     return modalView ? (
```
jeffmcaffer: suggest not using the ternary operator where the clauses are multiple lines as a "general" principle. Here you could just as well 

```javascript
if (modalView) return <Modal ...>
</Modal>
return <Grid ...>
</Grid>
```

src/styles/_FacetsEditor.scss
```diff
@@ -3,7 +3,8 @@
   padding-right: 0;
 }
 
-.facets__isEdited {
+.facets__isEdited,
```
jeffmcaffer: curious about the naming convention that uses `__`. How should this be interpreted vs camelCase and dashed-names? We currently have all three conventions.

src/utils/__tests__/contribution.test.js
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
```
jeffmcaffer: Awesome! tests!

src/utils/contribution.js
```diff
@@ -46,12 +56,27 @@ export default class Contribution {
     return result
   }
 
+  /**
+   * Build a Patch for a specific component
+   * @param {*} result initial object
+   * @param {*} component original component
```
jeffmcaffer: `changes` component spec?

src/utils/contribution.js
```diff
@@ -60,51 +85,228 @@ export default class Contribution {
...
+   * Returns the updated value or the original one if not modified
+   *
+   * @param  {Object} component original component
+   * @param  {Object} preview updated component sent back from the API after each change
+   * @param  {String} field field to check
    */
-  static getValue(component, changes, field) {
-    return changes && changes[field] ? changes[field] : this.getOriginalValue(component, field) || ''
+  static getValueAndIfDifferent(component, preview, field) {
+    const originalValues = get(component, field)
```
jeffmcaffer: consider

```javascript
static getValueAndIfDifferent(component, preview, field) {
    if (preview && get(preview, field)) return { value: get(preview, field), isDifferent: true }
    const originalValues = get(component, field)
    if (!originalValues || !originalValues.length) return [{ value: 'core', isDifferent: false }]
    return originalValues.map(value => ({ value: value || 'core', isDifferent: false }))
  }
```

Note that superficially not all cases return arrays. Not sure what is intendedn

src/utils/contribution.js
```diff
@@ -60,51 +85,228 @@ export default class Contribution {
...
+    return this.difference(preview, definition)
+  }
+
+  /**
+   * Deep diff between two objects, using lodash
+   * @param  {Object} object Object compared
+   * @param  {Object} base   Object to compare with
+   * @return {Object} Return a new object which represents the diff
+   */
+  static difference(object, base) {
```
jeffmcaffer: See if `deep-diff` can help here. we use it already in `CurationReview`. That code is largely dead given the new work you all are doing but the diffing is still relevant.

src/utils/contribution.js
```diff
@@ -60,51 +85,228 @@ export default class Contribution {
...
+
+  static isSourceComponent = component => ['github', 'sourcearchive'].includes(component.provider)
+
+  /**
+   *  Get image of definition based on the provider
+   *
+   * @param {*} item
+   * @returns image file
+   */
+  static getImage(item) {
```
jeffmcaffer: perhaps this should just be a table lookup?  providerImages?

src/utils/contribution.js
```diff
@@ -60,51 +85,228 @@ export default class Contribution {
...
+        return gem
+      case 'nuget':
+        return nuget
+      default:
+        return null
+    }
+  }
+
+  // Function that retrieve informations about facets from the definition
+  static foldFacets(definition, facets = null) {
```
jeffmcaffer: is this duplicate from DefinitionEntry? should be factored out

src/utils/definition.js
```diff
@@ -0,0 +1,25 @@
...
+// Abstract methods for Definition
+export default class Definition {
+  static getPathFromUrl(props) {
+    return props.path ? props.path : props.location ? props.location.pathname.slice(props.match.url.length + 1) : null
+  }
+
+  static getDefinitionPreview(state) {
+    return !state.ui.curate.previewDefinition.isFetching
+      ? Contribution.getChangesFromPreview(
+          Object.assign({}, state.ui.inspect.definition.item),
```
jeffmcaffer: do you want just shallow copy here? (not sure where this used to just an honest question)

src/api/clearlyDefined.js
```diff
@@ -9,10 +9,13 @@ export const API_DEVELOP = 'https://dev-api.clearlydefined.io'
 export const API_PROD = 'https://api.clearlydefined.io'
 export const apiHome = process.env.REACT_APP_SERVER || getServiceDefaultUrl()
 
+// Uncomment the next instruction to set your current enviroment as local, thus using API_LOCAL
+// DON'T COMMIT THIS FILE IF YOU CHANGED THIS INSTRUCTION
```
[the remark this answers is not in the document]
storrisi: We've done several tests trying to set process.env.NODE_ENV properly when starting the application, without success.
Given that, the options for a dev to switch enviroment are:
- set process.env.NODE_ENV in this file
- set API_LOCAL under the 'development' value of the switch.

Neither of them are the best way, but documenting the process maybe would be enough?

src/utils/contribution.js
AlexWebYourmind: Thanks for your comment, @jeffmcaffer. May I ask you to clarify which table would we be looking up? `providerImages` might be a new property in the schema or is something already available? 

Probably we are just a little bit unclear on this particular comment and it might be of great help if you could add additional details?


by the way, `getImage` is a very simple switch so maybe we can postpone this particular improvement at a later stage?

src/styles/_FacetsEditor.scss
AlexWebYourmind: @jeffmcaffer this is coming from the BEM naming conventions (http://getbem.com/naming/). 

The issue I see in the above code is that __ should be used for `block__elem` while `isEdited` is more of a modifier.

Hence, @storrisi  let's ensure to conver this (and any other modified) to `facets--isEdited` etc .

@jeffmcaffer can you confirm we are happy to use BEM?

src/api/clearlyDefined.js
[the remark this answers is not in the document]
AlexWebYourmind: I highly recommend the first one `set process.env.NODE_ENV` rather than changing back and forth the switch, @storrisi @jeffmcaffer .

Besides that, local configuration should generally be in .gitignore and not committed every time (other than the first time, where those act as a template).

CI/CD processes should populate the configuration accordingly (via e.g. SECRETS, etc).

Not sure which approach we are currently using on ClearlyDefined, but in this specific instance I think it might be beneficial to set process.env.NODE_ENV in the file.

Actually, an even better option, given we are using `create-react-app`, might be to create a .env file so that it will become possible to set process.env.NODE_ENV the right way (see https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-development-environment-variables-in-env)

src/components/FileList/FileList.js
storrisi: This current implementation could be used only if we maintain react-table.
Given that, we could sure evaluate if there's a way to simplify it, maybe into the way to keep some methods reusable when we will switch to ant-design/table.

src/styles/_FacetsEditor.scss
storrisi: I've fixed this, the right BEM convention is with double dash --

src/api/clearlyDefined.js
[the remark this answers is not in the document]
jeffmcaffer: The problem with process.env in CRA is that it has to be set at BUILD time, not run time. Since the CRA setup creates a static website there is code in there that does substitution etc. for NODE_ENV. I like the .env file approach. Again here the key point is that it is clear to the user 

* that there are multiple possible configurations
* how to get the different configs setup

src/utils/contribution.js
jeffmcaffer: I'm thinking of a new table in contribution.js (or elsewhere). It is more to get rid of the big (and ever growing) switch statement. This can be deferred if you like

src/styles/_FacetsEditor.scss
jeffmcaffer: I'm fine with BEM (don't know much about this space). One thing we should IMHO avoid is an explosion of "duplicate" styles. I've seen cases where there were 100s of styles with many basically being the same as one another but on a different page of the UI. It made it nearly impossible to get the UX consistent.

More generally, please do bring more coherence and consistency to the code! This would be a great next step once the current efforts for editing etc are done.

src/components/FileList/FileList.js
jeffmcaffer: OK for now. let's re-evaluate as we get further along. My more general comment is that large complex functions are an anti-pattern IMO. Love to see them broken down into bite-sized chunks that then fit together in a logical way the folks can understand easily at all levels of detail.

src/utils/contribution.js
storrisi: Thank you @jeffmcaffer , i needed to update the library since the recent version of deep-diff is 1.0.2 (the 0.3.8 that was in the package.json has some incompatibilities).
This saves me a lot of work / code.

src/utils/contribution.js
storrisi: Thank you, actually i needed to insert some improvements to this function, so i've take the chance to rewrite it in a more proper way

src/utils/contribution.js
jeffmcaffer: Awesome. Great to update the versions as we go as well.

src/api/clearlyDefined.js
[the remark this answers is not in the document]
AlexWebYourmind: Opened #257 to  address this

PR #259, opened 2018-08-24
Updated Facets and Globs rendering in Full Details View using ant-design tags

storrisi:
* updated FacetsRenderer using tags
* added GlobsPicker using tags
* updated FacetsEditor using GlobsPicker rather than InlineEditor


jeffmcaffer:
Looks good. Thanks

PR #262, opened 2018-08-27
adapt to the changes in source location

jeffmcaffer:
The new service code makes source location be much more like normal coordinates that happen to have a URL for convenience. This affects how it is output as well as how editing works


jeffmcaffer:
proactively merging to adapt to deployed service changes. Please review anyway.

PR #273, opened 2018-08-29
Various UI tweaks 

jeffmcaffer:
* remove some buttons and data from the Definitions list entries
* restructure the Full details view a bit
* update the scoring display on Definitions and full details

### DIFF
--- a/src/api/clearlyDefined.js
+++ b/src/api/clearlyDefined.js
@@ -75,13 +75,13 @@ export function previewDefinition(token, entity, curation) {
   return post(url(`${DEFINITIONS}/${entity.toPath()}`, { preview: true }), token, curation)
 }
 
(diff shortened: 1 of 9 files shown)


jeffmcaffer:
There is some quirk for how the Facet editor is laid out. seems that it is taking the full width of the page and then using col width 1 for the labels. I've not looked much to see why. That ends up making it hard to have the glob picker lined up in the right column. Not a huge deal.

Note that more data will be added to each section as we go.

jeffmcaffer:
There is also a quirk it seems like the first time the full details page is open (ever? on a given definition?) there is some error and the whole page reloads. Do it again and it works. It may be something with the network load speed (and then some value is null and we throw) or some other issue. Will need to investigate.

storrisi:
Looks good to me.

PR #286, opened 2018-09-01
Initial end-to-end test attempt using chrome + jest-puppeteer

mynameistechno:
First attempt at e2e tests 
* Uses chrome headless + puppeteer + jest. It also uses the jest preset `jest-puppeteer`: https://github.com/smooth-code/jest-puppeteer
* I really like the simplicity of the setup, how clean the tests are, and the speed of the tests (compared to selenium)
* I added a few sample tests for the Home page, Navbar, and the Definitions page
* To use:
  * `npm run e2e:start` to start a dev server at localhost:3333
  * `npm run e2e:test`
image

* We still need to sort out the backend for the integration tests. I.e. we need the service to serve a pristine data set for reproducible tests. Right now it currently uses a localhost instance that points to dev.clearlydefined.io. Ideally we would spin up the service in a docker container with some fixed data. Or we have a fake API service of sorts that we can generate off of swagger. @tesmart is looking into this.
* Note that I had to upgrade `react-scripts` to `@next` to get this to work as it needs a later version of jest. We just can't seem to get away from this 😝 However Babel 7 stable just announced (https://babeljs.io/blog/2018/08/27/7.0.0) and I believe `create-react-app` was in part waiting for babel 7 to release 2.0-stable 🤞 
* I'm not sure I want to commit to `jest-puppeteer` just yet. I can't seem to configure certain things (due to `create-react-app` interfering). I think I may just use `jest` + `puppeteer` directly and leverage the expect matchers provided by `jest-puppeteer`. We'll see.


jeffmcaffer:
Thanks @mynameistechno and @tesmart. What are the next steps here?

mynameistechno:
> Thanks @mynameistechno and @tesmart. What are the next steps here?
* This was more of an exploratory PR. I wanted to get thoughts from the other devs before committing. E.g. I could write the same tests using something like cucumber so we can compare. Thoughts?
* We need to sort out the service piece for this setup. We need pristine data for reproducible tests: 
  * We can fake API service responses (potentially easier to setup/less complexity, but needs to be updated whenever the service API is changed - perhaps swagger can help here though), or 
  * We spin up a full service instance with docker (more complexity to setup up initially, but then we have better integration tests that will help us find issues when something in the service changes and breaks the website), or
  * Currently the service API points to dev.clearlydefined.io. We could have a e2e.clearlydefined.io used only for e2e.
  * @tesmart was looking at this.  
* I'm waiting on this issue to be sorted out with jest-puppeteer, although we have a workaround for now with `e2e:start` https://github.com/smooth-code/jest-puppeteer/issues/123#issuecomment-418534343
*  This needed a recent version of `jest`, so I had to upgrade `react-scripts` to `@next` (alpha). So there's that. Like I mentioned above though, there is reason to believe it will be released soon seen Babel 7 was officially released.

jeffmcaffer:
@mynameistechno , @tesmart @storrisi how are we going to get to closure on this one. I'd very much like to get testing in this month. If we leave it too long then it will be a rush. who can drive the questions posed in @mynameistechno 's prior comment?

mynameistechno:
> @mynameistechno , @tesmart @storrisi how are we going to get to closure on this one. I'd very much like to get testing in this month. If we leave it too long then it will be a rush. who can drive the questions posed in @mynameistechno 's prior comment?

From our end, the big thing to sort out (regardless of which testing library we use) is the service piece of the setup. We should be able to work on this during our dedicated "open source friday" time. 

Other than that, if we stick with puppeteer, we need to harden the setup a bit so that we can integrate it into CI/Travis. I.e. one command to launch the website, service, run the tests, tear down, etc.

storrisi:
Thanks to @mynameistechno for his exploration about jest + puppeteer.
I've looked into the examples and those seems pretty easy to implement.
I'm addressing the addition of puppeteer into the website repo, and build the first tests.
We can close this.

jeffmcaffer:
I want to put this in but am nervous about the dependency on dev data. That changes all the time as people test different scnearios etc. We also want these tests to run for the prod deployment etc. I've not dug into the test structure. Is there a way to separate them out from the service? Can we stub that somehow?

mynameistechno:
> I want to put this in but am nervous about the dependency on dev data. That changes all the time as people test different scnearios etc. 

Right, that's the biggest thing left to sort before we can start pumping out a bunch of tests. Note that these e2e tests aren't triggered via `npm test`, they have a dedicated command that we can introduce to CI when we're ready.

> We also want these tests to run for the prod deployment etc. 

Yes they'd be great as a smoke test. One thing to consider though is that some tests will prob write data. So we can perhaps have a "smoke test script" with some simple read-only tests, or we have a way of tagging tests as read-only.

> I've not dug into the test structure. Is there a way to separate them out from the service? Can we stub that somehow?

The e2e tests work at the browser level, so you literally need to stand up a service endpoint of sorts for the website. @tesmart looked at this briefly. I.e. having a "fake" back-end with pre-canned json responses. It might work in the interim, but for true end-to-end/integration tests ideally we'd spin up the latest service instance, and load some controlled data. That way we'd know when a change introduced to either the website or the service breaks functionality. 

Maybe someone who is more familiar with the service side can suggest one way or the other? I.e. would it be easy to dockerize the service and have it seeded with some data?

AlexWebYourmind:
@mynameistechno let's keep in mind that e2e tests should clean after themselves, e.g. if we create real data as part of a "Curation" E2E test, there should be a set of cleanup tasks that will restore the situation to the previous status. 

Similarly, E2E tests should setup their own scenario (including running data creation steps so it will rely only on data created to test a specific scenario. E2E should be reproducible, so if we run it on a different environment it should provide an identical outcome, and not rely on existing data.

That said, there are ways to generate mock data for our tests, e.g. https://github.com/stympy/faker

mynameistechno:
> @mynameistechno let's keep in mind that e2e tests should clean after themselves, e.g. if we create real data as part of a "Curation" E2E test, there should be a set of cleanup tasks that will restore the situation to the previous status.

Sure, after every test you need to restore the database to it's previous state. My comment about read-only tests was in regards to running e2e smoke tests in production. That is, in most cases you don't want to run the full suite of e2e tests in production because restoring the prod db to its previous state is not always practical or worth the risk. 

> Similarly, E2E tests should setup their own scenario (including running data creation steps so it will rely only on data created to test a specific scenario. E2E should be reproducible, so if we run it on a different environment it should provide an identical outcome, and not rely on existing data.

Yes, they should set up their own data and be reproducible, that's what I mean when we need to sort out the service side of the e2e setup. However, the way we've done it in the past is that you spin up a clean environment for all the required services using docker containers. Running e2e on every environment might not be practical or be of much benefit. E.g. in a project at work, we have a terabyte database. Backing that up and restoring it after running tests is not feasible. We just use to docker to create a clean env and populate it with any needed data for the tests.

> That said, there are ways to generate mock data for our tests, e.g. https://www.npmjs.com/package/faker , but that would no longer be a proper E2E in my opinion. I prefer the Data generation / cleanup pattern better, maybe curating a dummy definition that is used only by the test itself. Faker (or similar) is the next best option anyways. (my humble opinion).

Yes, agreed. Long term we should hit a real instance of the service, if we went the faker route it would be just to get something going sooner than later.

jeffmcaffer:
Note that we can setup and run local servers that have local data etc. The service can all run with local (file system or in memory) stores so there is no need to run against production or even dev data. We could use the sample data repos as the test data or create new repos that have the data in them specifically for testing. then spin up a stand alone service using the data from these repos. We can also setup or stub out a crawler service to enable testing of the Harvest page end to end. (though realistically for UI testing we don't need to go that far)

This should cover everything except for GitHub interactions like creating PRs.

jeffmcaffer:
What is the status of this? @mynameistechno @AlexWebYourmind @storrisi 

Can we merge? Need to resolve the conflicts. See also #363

storrisi:
I've created a new PR #396 which address the E2E configuration based on this implementation.
I think we can close this.

fossygirl:
@storrisi @jeffmcaffer @AlexWebYourmind Okay, I'm closing this per the comment above, but please feel free to reopen if needed.

PR #301, opened 2018-09-14
Issue #243 - Chunk up the definition fetching

### LINKED ISSUE #243: Chunk up the definition fetching
I dropped a package-lock file (1145 packages) on the browse window and it took 1.4 minutes for the defintions to come back. since the user is looking at the first set, and the service itself is likely just doing a lot of IO, we should chunk this up into a set of requests of say 100 coordinates at a time. Consider actually enforcing this on the service as way of forcing ratelimiting.

storrisi:
Chunk up the definition fetching by 100, fixes #243

### DIFF
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -104,18 +104,18 @@ class PageDefinitions extends AbstractPageDefinitions {
   refresh = () => {
     const { components, dispatch } = this.props
 
-    this.onRemoveAll()
     const definitions = this.buildSaveSpec(components.list)
+    const definitionsToGet = definitions.map(definition => definition.toPath())
 
-    const definitionsToGet = []
-    definitions.map(definition => {
-      definitionsToGet.push(definition.toPath())
-      delete definition.changes
-    })
+    if (this.hasChanges()) {
+      dispatch(
+        uiBrowseUpdateList({
+          transform: list => list.map(({ changes, ...keepAttrs }) => keepAttrs)
+        })
+      )
+    }
 
     this.getDefinitionsAndNotify(definitionsToGet, 'All components have been refreshed')
-
-    dispatch(uiBrowseUpdateList({ addAll: definitions }))
   }
 
   renderButtons() {
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -10,6 +10,7 @@ import base64js from 'base64-js'
 import { saveAs } from 'file-saver'
 import notification from 'antd/lib/notification'
 import AntdButton from 'antd/lib/button'
+import chunk from 'lodash/chunk'
 import { FilterBar } from './'
 import { uiNavigation, uiBrowseUpdateList, uiNotificationNew } from '../actions/ui'
 import { getDefinitionsAction } from '../actions/definitionActions'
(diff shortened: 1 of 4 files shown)


src/components/PageDefinitions.js
```diff
@@ -87,20 +88,35 @@ class PageDefinitions extends AbstractPageDefinitions {
...
   }
 
+  // Get an array of definitions asynchronous, split them into 100 chunks and alert the user when they're all done
+  getDefinitionsAndNotify(definitions, message) {
+    const { dispatch, token } = this.props
+    const chunks = chunk(definitions, 100)
+    let todo = chunks.length
+    for (let i = 0; i < chunks.length; i++) {
+      dispatch(getDefinitionsAction(token, chunks[i])).then(() => {
+        if (--todo === 0) {
```
jeffmcaffer: feels like this should be more like 

```javascript
Promise.all(chunks.map(chunk => dispatch(getDefinitionAction(...))).then(()=> notify)
```

likely should have a error case where we tell the user that something bad happened. Perhaps the caller can pass in a failure message?

src/components/PageDefinitions.js
```diff
@@ -87,20 +88,35 @@ class PageDefinitions extends AbstractPageDefinitions {
...
+    const { components, dispatch } = this.props
 
     this.onRemoveAll()
     const definitions = this.buildSaveSpec(components.list)
 
-    definitions.forEach(definition => {
-      const path = definition.toPath()
+    const definitionsToGet = []
+    definitions.map(definition => {
+      definitionsToGet.push(definition.toPath())
```
jeffmcaffer: I'd like to review how refresh works from a user point of view. Ideally the data just stays there and is updated if need be. does `onRemoveAll` cause all the data to disappear? If we are really throwing away the definitions then the `delete` should not be needed.

In an ideal future state, we don't need to throw away the user's changes if there are no changes in the underlying data.

gianpaj:
Hi @jeffmcaffer, thanks for the feedback. It's true that the list components was being removed and quickly added back. 
Now the list is updated, only if necessary

src/components/PageDefinitions.js
```diff
@@ -87,20 +88,34 @@ class PageDefinitions extends AbstractPageDefinitions {
     }
   }
 
+  // Get an array of definitions asynchronous, split them into 100 chunks and alert the user when they're all done
+  getDefinitionsAndNotify(definitions, message) {
+    const { dispatch, token } = this.props
+    const chunks = chunk(definitions, 100)
+    Promise.all(chunks.map(chunk => dispatch(getDefinitionsAction(token, chunk))))
```
jeffmcaffer: Might consider using `throat(10)` or something here to not abuse the service. If the UI is showing 2000 definitions then refresh would do 20 parallel 100 element requests. Just be friendly.

src/components/PageDefinitions.js
```diff
@@ -87,20 +88,34 @@ class PageDefinitions extends AbstractPageDefinitions {
...
+        )
+      )
+  }
+
   refresh = () => {
-    const { components, dispatch, token } = this.props
+    const { components, dispatch } = this.props
 
-    this.onRemoveAll()
     const definitions = this.buildSaveSpec(components.list)
```
jeffmcaffer: nit: move these two lines down to L118 where the values are used.

src/components/PageDefinitions.js
```diff
@@ -243,16 +258,17 @@ class PageDefinitions extends AbstractPageDefinitions {
...
 
-  async loadFromListSpec(listSpec) {
-    const { dispatch, token, definitions } = this.props
+  loadFromListSpec(listSpec) {
+    const { dispatch, definitions } = this.props
     if (listSpec.filter) this.setState({ activeFilters: listSpec.filter })
     if (listSpec.sortBy) this.setState({ activeSort: listSpec.sortBy })
     if (listSpec.sortBy || listSpec.filter) this.setState({ sequence: this.state.sequence + 1 })
 
-    const toAdd = listSpec.coordinates.map(component => EntitySpec.validateAndCreate(component)).filter(e => e)
```
jeffmcaffer: don't need to filter nulls?

gianpaj: I didn't know there could be nulls here. I've added it back

src/components/PageDefinitions.js
gianpaj: Sounds fair. I've done that

gianpaj:
Hi @jeffmcaffer, I've made the changes you suggested. Please take a look

PR #308, opened 2018-09-21
Issue #197 - Revert individual or all changes to contributions made

### LINKED ISSUE #197: Revert individual or all changes to contributions made
Everywhere that the user is able to propose a change (e.g., expanded or full view) they can also revert individual changes as well as all changes.

I the case of the definition list, the user can revert an individual change, all changes for an individual definition or all changes in the list. (note that if we end up doing multi-select, then reverting all changes in the selection would be a good option too)

storrisi:
Fixes #197

### DIFF
--- a/src/actions/definitionActions.js
+++ b/src/actions/definitionActions.js
@@ -3,6 +3,8 @@
 
 import { asyncActions } from './'
 import { getDefinitions, getDefinition, previewDefinition, getDefinitionSuggestions } from '../api/clearlyDefined'
+import Definition from '../utils/definition'
+import { uiBrowseUpdateList } from './ui'
 
 export const DEFINITION_LIST = 'DEFINITION_LIST'
 export const DEFINITION_BODIES = 'DEFINITION_BODIES'
@@ -60,5 +62,16 @@ export function resetPreviewDefinitionAction(token, entity, name) {
     const actions = asyncActions(name)
     dispatch(actions.start())
     return dispatch(actions.success({}))
+  }
+}
+
+export function revertDefinitionAction(definition, values, name) {
+  return (dispatch, getState) => {
+    const state = getState()
+    const actions = asyncActions(name)
+    dispatch(actions.start({ definition, values }))
+    const componentsWithoutChanges = Definition.revert(state.ui.browse.componentList.list, definition, values)
+    dispatch(uiBrowseUpdateList({ updateAll: componentsWithoutChanges }))
+    return dispatch(actions.success({ componentsWithoutChanges }))
   }
 }
--- a/src/components/FullDetailView/FullDetailPage.js
+++ b/src/components/FullDetailView/FullDetailPage.js
@@ -5,6 +5,7 @@ import React, { Component, Fragment } from 'react'
 import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
 import { Grid, Button } from 'react-bootstrap'
+import omitBy from 'lodash/omitBy'
 import isEmpty from 'lodash/isEmpty'
 import cloneDeep from 'lodash/cloneDeep'
 import PropTypes from 'prop-types'
@@ -18,7 +19,8 @@ import {
   uiInspectGetHarvested,
   uiNavigation,
   uiCurateGetDefinitionPreview,
-  uiCurateResetDefinitionPreview
+  uiCurateResetDefinitionPreview,
+  uiRevertDefinition
 } from '../../actions/ui'
 import { ROUTE_DEFINITIONS } from '../../utils/routingConstants'
 import { curateAction } from '../../actions/curationActions'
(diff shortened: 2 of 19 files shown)


src/components/FullDetailView/FullDetailPage.js
```diff
@@ -165,6 +171,45 @@ export class FullDetailPage extends Component {
...
+    const { uiCurateResetDefinitionPreview } = this.props
+    const { changes } = this.state
+    if (isEmpty(changes)) return
+    if (value) {
+      const revertedChanges = omitBy(changes, (_, index) => index.startsWith(value))
+      this.setState({ changes: revertedChanges, sequence: this.state.sequence + 1 }, () => this.previewDefinition())
+      return
+    }
+    const key = `open${Date.now()}`
+    const NotificationButtons = (
```
jeffmcaffer: feels like this should be some reuseable notification/confirmer or at least a separate function (renderConfirm or some such)

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -165,6 +171,45 @@ export class FullDetailPage extends Component {
...
+          }
+        >
+          Confirm
+        </AntdButton>
+        <AntdButton type="secondary" size="small" onClick={() => notification.close(key)}>
+          Dismiss Notification
+        </AntdButton>
+      </Fragment>
+    )
+    notification.open({
```
jeffmcaffer: still a bit fussed about this pattern of `open()` rather than having a prop like `show: true`. Same for `close`

src/actions/definitionActions.js
```diff
@@ -59,3 +61,14 @@ export function resetPreviewDefinitionAction(token, entity, name) {
     return dispatch(actions.success({}))
   }
 }
+
+export function revertDefinitionAction(definition, values, name) {
+  return (dispatch, getState) => {
+    const state = getState()
+    const actions = asyncActions(name)
+    dispatch(actions.start({ definition, values }))
+    const componentsWithoutChanges = Definition.revert(state.ui.browse.componentList.list, definition, values)
```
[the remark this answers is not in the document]
storrisi: I got the point, i will change this naming to something more clear

src/components/FullDetailView/FullDetailPage.js
storrisi: The [Notification component](https://ant.design/components/notification/) in antd-design is actually an API. It requires to call those methods (open and close) in order to show or hide the notification

src/components/FullDetailView/FullDetailPage.js
storrisi: Makes sense, we could definitely improve this.
I will commit a solution into this PR

src/utils/definition.js
```diff
@@ -30,4 +31,32 @@ export default class Definition {
...
+    })
+  }
+
+  static revertChanges(component, value) {
+    if (value) {
+      const { [value]: omit, ...withoutChanges } = component.changes
+      component.changes = withoutChanges
+      return component
+    } else {
+      const { changes, ...withoutChanges } = component
```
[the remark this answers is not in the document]
storrisi: The goal here is actually to preserve the initial object, instead of deleting property from it.
Given that, i found the spread operator + destructuring a much more effective solution.

I made some tests before posting an aswer.
Actually, this implementation of spread operator + destructuring, doesn't mutate the original object.
`const { changes, ...withoutChanges } = component` creates a new object `withoutChanges` which contains everything from `component` except the property `changes`.

I've put 2 console.log in the code as proof:
schermata 2018-09-26 alle 11 24 58
and this is the result:
schermata 2018-09-26 alle 11 24 47

As you can see, the component object has not been mutated.

Since the purpose is to not change the values in the components object, but instead return a new object and then using it to update the redux store, this would not affect any other use of the components object.

Makes sense?

storrisi:
@jeffmcaffer I've made some updates based on your suggestions.
You can now review again the PR.

src/utils/definition.js
```diff
@@ -30,4 +31,33 @@ export default class Definition {
...
+  /**
+   * Revert a list of definitions or a specific one, removing all the changes or only specific values
+   * @param  {[]} components list of definitions
+   * @param  {{}} definition specific definition, if null the function will check all the components
+   * @param  {string} key string that identifies the specific value to revert, if null all the changes will be removed
+   */
+  static revert(components, definition, key) {
+    if (!components) return
+    return components.map(component => {
+      const updatedChanges = this.revertChanges(component, key)
```
jeffmcaffer: This should only run if needed. More to line 49

src/utils/definition.js
```diff
@@ -30,4 +31,33 @@ export default class Definition {
...
+        return updatedChanges
+      else return component
+    })
+  }
+
+  static revertChanges(component, key) {
+    if (key) {
+      const { [key]: omit, ...updatedChanges } = component.changes
+      return { ...component, changes: updatedChanges }
+    } else {
```
jeffmcaffer: Redundant else can be removed

src/components/PageDefinitions.js
```diff
@@ -87,8 +84,21 @@ class PageDefinitions extends AbstractPageDefinitions {
     }
   }
 
+  // Get an array of definitions asynchronous, split them into 100 chunks and alert the user when they're all done
+  getDefinitionsAndNotify(definitions, message) {
+    const { dispatch, token } = this.props
+    const chunks = chunk(definitions, 100)
+    Promise.all(chunks.map(throat(10, chunk => dispatch(getDefinitionsAction(token, chunk)))))
```
jeffmcaffer: Using try/ catch and async/ await here is likely better. More in keeping with the rest of the code and safer.  As is, if some thing in the map fails before returning the promise, the catch handler will not catch it.  That failure would happen while committing the arg to all and would not reject a promise.

jeffmcaffer:
@storrisi Note that I merged but there is still an unaddressed comment relating to the error handling in `getDefinitionsAndNotify`. I think it should be using a `try/catch` with `async/await` structure.

storrisi:
@jeffmcaffer thanks, i was actually working on that yesterday. I will address this in another PR.

Issue #168, opened 2018-06-20
Add Gist support for saving of lists


fossygirl:
We should add Gist support to our save functionality for lists: users should be able to have Gists as a mechanism for storing exported lists and they should be able to drop a Gist URL into the definition list and have the content of the Gist read and the corresponding definitions loaded.

caabernathy:
Just posting my prelim investigations. The underlying component handling Drag and Drop (react-dropzone) doesn't currently handle dragging in URLs. Different from dragging in files, the underlying code rejects the process because the "file name" is `undefined`. It's undefined because the dropzone component is actually an `<input type="file" />` under the hood.

There may be some additional hoops needed to make this work but it isn't obvious to me how this can be done. 

Perhaps, the input's `value` attribute or other element can be extracted to get at the URL and then fetch that content when the default path fails.

Maybe there's a different react component that can handle dragging different type of input other than files. I did a search and haven't found anything thus far.

storrisi:
Given that now that DnD behaviour is implemented, we can work on this.
We can use the Gist APIs to save on a newly Gist file, or retrieve the contents from a dropped Gist URL.

I assume that the content of the file should be the same of the json export.
@fossygirl is that right?

PR #310, opened 2018-09-21
Issue #136 - constrain the type of files that can be dropped

### LINKED ISSUE #136: constrain the type of file that can be dropped
The current drop zone on the Browse page allows pretty much any file to be dropped. This should be constrained to only the types of files we understand (e..g, by file name and extension if possible). If the file we accept is somehow not correct then a more detailed message should pop up.

storrisi:
Fixes #136

### DIFF
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -198,6 +198,7 @@ class PageDefinitions extends AbstractPageDefinitions {
 
   onDrop(acceptedFiles, rejectedFiles) {
     const { dispatch } = this.props
+    if (!acceptedFiles.length) return
     dispatch(uiNotificationNew({ type: 'info', message: 'Loading component list from file(s)', timeout: 5000 }))
     acceptedFiles.forEach(file => {
       const reader = new FileReader()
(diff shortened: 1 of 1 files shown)


src/components/PageDefinitions.js
```diff
@@ -213,6 +214,11 @@ class PageDefinitions extends AbstractPageDefinitions {
     })
   }
 
+  onDropRejected = files => {
+    const fileNames = files.map(file => file.name).join(', ')
+    this.props.dispatch(uiNotificationNew({ type: 'danger', message: `Could not load: ${fileNames}`, timeout: 5000 }))
```
jeffmcaffer: The message should be more informative. It is not that we tried to load and could not it is that the type of the file is not supported.

src/components/PageDefinitions.js
```diff
@@ -223,7 +229,13 @@ class PageDefinitions extends AbstractPageDefinitions {
 
   dropZone(child) {
     return (
-      <Dropzone disableClick onDrop={this.onDrop} style={{ position: 'relative' }}>
+      <Dropzone
+        accept="application/json"
```
jeffmcaffer: how will this work when we start to accept lots more types of files? Right now it is just json cause that was simple and easy but more generally we will be taking just about anything we can understand. It's ok for now this way but I'm interested to understand the future.

Note also that we need to be able to drop things other than files at some point. In particular, if someone drags a URL from a browser.

storrisi: This needs of course to be improved, there are already some opened issues about DnD and the ability to allow any kind of accettable data.
We would end with a specific utils class method that return the allowed file formats, and of course with a related method that would check the format of those files.
For DnD we need to understand if DropZone is able to manage them, or if we should use something different (see comments on #168 )

PR #311, opened 2018-09-21
Issue #162 - High level view of definition (icon and empty component)

### LINKED ISSUE #162: High level view of definitions
Adding a component to the list should show users a high level view of the definition that includes:

- indications of type, provider, namespace, name, and version
- ClearlyDefined score
- Declared SPDX License expression
- Indicators as to whether or not the entry has any data at all in the system, been curated, or has outstanding contributions

storrisi:
Fixes #162 , completing the outstanding implementation for that issue

### DIFF
--- a/src/components/ComponentList.js
+++ b/src/components/ComponentList.js
@@ -109,7 +109,7 @@ export default class ComponentList extends React.Component {
     const { readOnly, hasChange } = this.props
     const isSourceComponent = this.isSourceComponent(component)
     const scores = Definition.computeScores(definition)
-    const isDisabled = Definition.isDefinitionEmpty(definition)
+    const isSourceEmpty = Definition.isSourceEmpty(definition)
     return (
       <div className="list-activity-area">
         {scores && <img className="list-buttons" src={getBadgeUrl(scores.tool, scores.effective)} alt="score" />}
@@ -122,17 +122,18 @@ export default class ComponentList extends React.Component {
               </Button>,
               'Add the definition for source that matches this package'
             )}
-          {this.renderButtonWithTip(
-            <Button
-              className="list-fa-button"
-              onClick={this.inspectComponent.bind(this, currentComponent, definition)}
-              disabled={isDisabled}
-            >
-              <i className="fas fa-search" />
-            </Button>,
-            'Dig into this definition'
-          )}
+          {!isSourceEmpty &&
+            this.renderButtonWithTip(
+              <Button
+                className="list-fa-button"
+                onClick={this.inspectComponent.bind(this, currentComponent, definition)}
+              >
+                <i className="fas fa-search" />
+              </Button>,
+              'Dig into this definition'
+            )}
           {!readOnly &&
+            !isSourceEmpty &&
             this.renderButtonWithTip(
               <Button
                 className="list-fa-button"
--- a/src/components/ComponentList.js
+++ b/src/components/ComponentList.js
@@ -109,6 +109,7 @@ export default class ComponentList extends React.Component {
     const { readOnly, hasChange } = this.props
     const isSourceComponent = this.isSourceComponent(component)
     const scores = Definition.computeScores(definition)
+    const isDefinitionEmpty = Definition.isDefinitionEmpty(definition)
     const isSourceEmpty = Definition.isSourceEmpty(definition)
     return (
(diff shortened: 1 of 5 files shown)


src/components/ComponentList.js
```diff
@@ -104,25 +104,35 @@ export default class ComponentList extends React.Component {
...
+            !readOnly &&
+            this.renderButtonWithTip(
+              <Button onClick={this.addSourceForComponent.bind(this, component)}>
+                <i className="fas fa-code" />
+              </Button>,
+              'Add the definition for source that matches this package'
             )}
           {this.renderButtonWithTip(
-            <Button className="list-fa-button" onClick={this.inspectComponent.bind(this, currentComponent, definition)}>
+            <Button
```
jeffmcaffer: better here would be to only show the "add source" button if there is actually a sourcelocation in the given definition. That works for the empty case as well as improving the experience when there is a definition and it just happens to not have a source location.

Also means (I think) that you can get rid of the whole "disabled" concept.

src/components/TwoLineEntry.js
```diff
@@ -13,18 +13,19 @@ export default class TwoLineEntry extends React.Component {
...
     onClick: () => {}
   }
 
   render() {
-    const { buttons, image, headline, message, onClick, letter, panel, highlight } = this.props
+    const { buttons, image, headline, message, onClick, letter, panel, highlight, isDisabled } = this.props
     return (
       <div className="two-line-entry">
-        <div className="list-row" onClick={onClick}>
+        <div className={`list-row${isDisabled ? ' isDisabled' : ''}`} onClick={onClick}>
```
jeffmcaffer: suggest we change from "disabled" to "empty" or some such. We might actually want the user to be able to do some things on empty entries (e.g., in theory they could start adding curations)

src/utils/definition.js
```diff
@@ -30,4 +30,14 @@ export default class Definition {
...
     return { tool, effective }
   }
+
+  /**
+   * Determine if a component doesn't have any data. In order to show in the UI in the list of Components
+   *
+   * @param {*} definition
+   * @returns {boolean}
+   */
+  static isDefinitionEmpty(definition) {
```
jeffmcaffer: This naming and commenting is too broad. There could be tons of other data, and just the source location is missing (actually that is the 90% case).

storrisi:
@jeffmcaffer this is how the implementation looks like:
sep-21-2018 17-28-58

src/components/ComponentList.js
storrisi: @jeffmcaffer i've updated the code based on your suggestion.

This is what is gonna happen if the definition has an empty sourceLocation:
schermata 2018-09-28 alle 09 31 19

storrisi: @jeffmcaffer what should be the best way to check if a definition is empty?

Actually, a definition that hasn't any data, has a structure like this:
schermata 2018-09-28 alle 15 30 51

Given that, we checked for the existence of the field "sourceLocation".
That's why, from our perspective, the best way to check an empty definition was to check the sourceLocation field.

storrisi:
I don't know why you can't see my comments.
By the way i've improved the method that check if a definition is empty, watching for the `described.tools` array field

src/utils/definition.js
```diff
@@ -32,6 +33,20 @@ export default class Definition {
...
+    return !definition.described || (definition.described && isEmpty(definition.described.tools))
+  }
+
+  static isSourceEmpty(definition) {
+    return !definition.described || (definition.described && definition.described.sourceLocation === undefined)
```
jeffmcaffer: Simplify to `return !!get(definition, 'described.sourceLocation')`

src/utils/definition.js
```diff
@@ -32,6 +33,20 @@ export default class Definition {
...
   }
 
+  /**
+   * Determine if a component doesn't have any data. In order to show in the UI in the list of Components
+   *
+   * @param {*} definition
+   * @returns {boolean}
+   */
+  static isDefinitionEmpty(definition) {
+    return !definition.described || (definition.described && isEmpty(definition.described.tools))
```
jeffmcaffer: simplify. `return isEmpty(get(definition, 'described.tools'))`

Issue #289, opened 2018-09-03
Structure for standalone components


jeffmcaffer:
Many of the UI components we are building (e.g., license picker, source location picker, ...) are likely interesting to others. Certainly people looking to embed ClearlyDefined flows into their products or flows would find it useful to be able to reuse our components.

Here we should consider the options for both development and delivery. 
1. One repo with many components producing one npm
1. One repo with many components producing multiple npms
1. Separate repos for each component producing separate npms

If we # 2 there will also be the question of which mono-repo structure do we use (e.g., rush, lerna, ...)

AlexWebYourmind:
@jeffmcaffer Thanks for the additional information.

To clarify, the deliverables for this issues are:

- New CD Repo
- Initial NPM Package Deployment
- Documentation  (e.g. Readme.MD outlining the structure, how to use it etc)

Can you confirm the above is correct? 

We can then open separated issues to e.g. make an existing component standalone (if not already) + add to new repo.

Thoughts?

jeffmcaffer:
@AlexWebYourmind that is more or less right. I was chatting with @valtterihalla about #179 and thought that them adding the new source location picker would be a good opportunity to scope out a design for this new component structure. We can them move over any existing generic components as needed.

Hopefully @valtterihalla and team can put together some proposals and we can iterate and then implement.

daniellandau:
Is the idea that users of components have to provide their own API endpoints for example for Github search, or should they call endpoints in ClearlyDefined? The ClearlyDefined API already does CORS allowing all origins so in principle e.g. `GithubSelector.js` and `GithubCommitPicker.js` are publishable almost as is, as they do not plug in to any redux state management and their only dependency to anywhere else in the project is to get the URL to shoot requests at.

These end points in the API are wide open to the whole world already, but of course exposing easy React components to NPM would probably increase their usage. This might then be either a good or a bad thing, depending on what kind of agreement we can get on Github (and other, Gitlab, Maven, NPM, etc) quotas.

That's for read-only and external-only API calls. What about actual ClearlyDefined data and embeddable components contributing back to ClearlyDefined? I think for those it would be quite necessary to just directly call `api.clearlydefined.io`.

Maybe we could just do all the calls to `api.clearlydefined.io` but really clearly mark the NPM packages as experimental with no guarantees of continuing access to the API.

jeffmcaffer:
[...]

PR #313, opened 2018-09-27
[PROPOSAL] Extract components to independent NPM package

daniellandau:
This is a proposal for https://github.com/clearlydefined/website/issues/289

I decided not to publish to NPM at this stage, so there is something temporary in the code, namely this in `package.json`:

```     "clearlydefined-components": "file:components/clearlydefined-components-0.1.0.tgz", ```

In the end that would just be `"^0.1.0"` or whichever version is current on the right side. For the time being to test this, one needs to first run `npm install && npm pack` in `components/` before running `npm install` in the root.

### CODE REVIEW

**Comment by daniellandau** [comment]
That Travis fail is expected and should be fixed once the NPM package is published and the ref changed to that.

**Comment by jeffmcaffer** [comment]
Thanks @daniellandau. did you look at things like https://www.npmjs.com/package/@microsoft/rush or https://github.com/lerna/lerna?

**Comment by jeffmcaffer** [comment]
In the end the will be many of the plugins. Pretty much each of the different language sources will be at least one other extension.  We want to make distinct packages for many of the react components so they can be used in other contexts and so it's easy for others to contribute.

**Comment by daniellandau** [comment]
I understood from https://github.com/clearlydefined/website/issues/289#issuecomment-418156361 that you explicitly wanted to go with a single NPM package. What do you "each of the different language sources will be at least one other extension."?

Having different React components in the same package shouldn't be a big problem, as first of all the files are small, and second users with tree shaking bundlers (such as webpack 2 used in create-react-app) will not be exporting unused components to their users.

**Comment by daniellandau** [comment]
@jeffmcaffer the file git conflict is now resolved.

**Comment by jeffmcaffer** [comment]
@daniellandau The travis build is failing because of the components folder changes (it is expecting a packed npm in the file system). In the end devs should not have to pack before running. This messes with the development experience. Ideally we just `npm link ./components` and go.

I created #324 as a working place for me to try things. Feel free to pull bits from there if you think it is useful. 

In the end the devs should only have to clone, run npm install at the root and in the ./components dir, and then run npm link ./components. From there they should run. I have verified that that works with the changes in my branch. The trick is getting travis to do the right thing.

**Comment by daniellandau** [comment]
@jeffmcaffer this is ready for review again.

**Comment by daniellandau** [comment]
Note that when merging, a new version of ui-components needs to be published as well as the API has changed.

File: components/src/GitHubCommitPicker.js [hunk 1]
```diff
@@ -6,11 +6,28 @@ import PropTypes from 'prop-types'
 import { Typeahead, Highlighter } from 'react-bootstrap-typeahead'
```
storrisi: I'm not so sure if is fair enough to force a developer to use a specific library for styling.
If we want to release some stand-alone components, perhaps it could be better to give to the user the basic functionality without any styling implementation, and neither third-party libraries if not necessary.

Stand-alone components should be clean as much as possible, then each developer can improve the styling as he prefers.
@jeffmcaffer thoughts?

storrisi:
Overall i agree with the implementation.
Seems to be smart and simple enough to manage and let the developers improve the library as well.
I've pointed out only a few questions about including third-party libraries in the components, specially ui-libraries.
Besides this, i think we can move further with this and start to identify which are the components that should be rendered as stand-alone, and remove from them any specific implementation and dependency used in this project.

components/src/GitHubCommitPicker.js
jeffmcaffer: It's a good point. We could allow folks to plug in a component to use for various cases. If that is relatively straight forward, sure. We could also do this based on demand. If people want different styling then we can adapt. I propose that we factor out a representative set of components and get a sense of the overall styling needs and build up a more holistic approach.

AlexWebYourmind:
@jeffmcaffer @daniellandau  What's the latest on this particular PR?  I have noticed branch conflicts. If we are happy with the suggested implementation we should resolve the conflicts and merge the changes?

daniellandau:
It was discussed earlier that a separate repo most probably doesn't bring enough added value to offset the complexity cost. Getting this merged would bring benefit though. I'd be happy to resolve the conflicts once I have indication that it will be merged soonish after that. @jeffmcaffer?

storrisi:
@daniellandau I've tried this solution in a local environment while trying to convert a component in a standalone one.

These are the issues that i've encountered:
- after running `npm install` from the `components` folder, and ran an `npm link ./component` from the website folder, seems like that some libs are still missing. So i've ended up running an `npm build` from the components directory
- that last operation has some babel issues since it didn't recognize some js intructions, which failed the build
- an npm `@clearlydefined/ui-components` exists, so if the user try to run a simple `npm install` from the website folder, without make the link before, the system try to install the package from the remote one, which has a different version from the local one, causing an error while installing

jeffmcaffer:
I'm not sure what to do here. There does not seem to be consensus on how to achieve the goal of one repo producing multiple npms and yet still being able to do reasonable local development. I feel like we are falling into the pit of quirks that things like Rush and Lerna are here to address. 

We should reassess the priority of this outcome (I very much want it) relative to actual demand

storrisi:
@jeffmcaffer i've tested this in #409 while trying to render the License Picker as standalone.
If configured in the right way, this proposal could work, but maybe it's not properly a standard and somehow feels like reinventing the wheel if compared to Lerna.
As i said in #409 i'd be happy to give Lerna a try.

jeffmcaffer:
This one has been dangling for quite a while. I'm uncomfortable putting together our own solution here as it is a very complicated and subtle space and people have already spent huge amounts of effort figuring it out. We could so something that works today for the UI but it would be better to head in a direction that could work for all the different repos or even merging all the repos into one in the future.

@storrisi did a quick poke ant Lerna and it was promising. It is a pretty fundamental change though so we should do it deliberately. 

Closing this PR but the related issue #289 will still track the work. I'll add comments there similar to the above. Thanks @daniellandau for the effort here. If you want to jump in on the Lerna direction, that would be great.

Issue #291, opened 2018-09-03
Apply related curations


jeffmcaffer:
It often happens that one curation can apply to multiple components. For example, a missing license on v1.0 might be applicable to v1.1, 1.2, 2.0, ... We do not want to automate that but we should make it very easy to apply an existing curation to related components. 

For example, when looking at v2.0 the user should be told that here are curations to other versions. Note that it may got either way. Could be that I curated 1.0 and that could apply to 2.0 or vice versa. Just depends on the order of curation.

Consider pulling in non-curated data as well. For example, say 1.0 does not have a license and no curation. V1.1 they got smart and added a license. We can highlight that in 1.0 and "propose" that as a curation for the user to confirm and submit.

Be smart and don't allow pulling over things that clearly do not apply, like the commit hash.

Issue #34, opened 2018-01-22
Hook up Add Curation button


jeffmcaffer:
on /components there is an `Add curation` button. Currently it is not connected to anything. It should open the /curate page on the component currently being viewed.

PR #316, opened 2018-09-28
WIP: Issue #291 - Apply Related Curations (#34)

storrisi:
Related to #291 

This is the current implementation:
set-29-2018 00-06-45

As you can see, the UI shows to the user the highlighted suggestions from the last curation of the definition (despite the revision).
Then the user is able to keep or discard each suggestion, and provide a new Contribution with those data.

Data showed to the user are currently provided by a mocked API

This PR contains some refactor of the FullDetailComponent, which is actually quite long (>378 lines of code before this PR)


jeffmcaffer:
Looks good from a user point of view. I've not looked at the code yet. 

As I was thinking this through I wondered if the metaphor here should be "other revisions of this component had these values for this property" sort of like Amazon suggestions.
* Does not matter if the suggested value comes from a curation. Could come from a tool run on a more clearly defined version (earlier or later)
* Even if there is a value in the current one, might be interesting to point out that other versions had a different value
* Avoids the problem of when to "suggest". effectively we always show the means of seeing the other values. These could just be in the "pickers" for the values. For example. have a "related values" section in the license picker. User could see that related components have other licenses and maybe even navigate to those definitions (separate tab).

Some comments:

* Will we be able to do the same thing in the drop down panel?
* Will there be some sort of indication on the two line and/or panel view that there are suggestions? Or will the user need to ask for them?
* Just because I noticed you show it, not sure how we can do the source location as a suggestion unless it is using the GitHub repo and finding a tag that matches the current version. That is, the reusable part of a related source location curation is the repo info, not the `revision`.
* Are you thinking that individual files would also have these affordances?

storrisi:
> Does not matter if the suggested value comes from a curation. Could come from a tool run on a more clearly defined version (earlier or later)

I guess those data are returned from other revisions of the same definition.
Looking at the proposed API from your document, probably is actually missing an endpoint that returns definitions data for any revision?

> Even if there is a value in the current one, might be interesting to point out that other versions had a different value
> Avoids the problem of when to "suggest". effectively we always show the means of seeing the other values. These could just be in the "pickers" for the values. For example. have a "related values" section in the license picker. User could see that related components have other licenses and maybe even navigate to those definitions (separate tab). 

For these 2 point we could end up with a solution similar to the one Facebook use for Ads Management.
The Suggestion label is highlighted only if there is at least one suggestion. Could this proposal be a good solution?
schermata 2018-10-01 alle 12 19 03

> Will we be able to do the same thing in the drop down panel?

Is the drop down panel actually the Components List?
In that case, we have thought to highlight something to the user, in order to only let know that there are suggestions.
I would not let the user check the suggestions directly from the list, since it could mean to make a lot of calls for each definition in the list, ending with a huge time waiting if the list is pretty long.
What do you think about it?

> Will there be some sort of indication on the two line and/or panel view that there are suggestions? Or will the user need to ask for them?

As mentioned above, we think the best way is to let the user know that a definition has some suggested data, and from the two line list inspect the full details view to see what are the suggested changes.
Likewise the user, in a full details view, should to be notified that there are suggestions (especially if he's checking from a deep linked URL)
This was originally the reason of the highlighted field.
But, given that the user could make a choice, then i would only suggest to highlight the field, and then let the user see all the suggestions in the proposed way.

> Just because I noticed you show it, not sure how we can do the source location as a suggestion unless it is using the GitHub repo and finding a tag that matches the current version. That is, the reusable part of a related source location curation is the repo info, not the revision.

Since there's a new proposal implementation for the Source Location Picker, should we leave this for now?

> Are you thinking that individual files would also have these affordances?

They should, given that any file could have license, facets or copyrights updates.
On the other way, it could very uncomfortable for the user to check any single change for any file, especially if there a lot of files.
Like we've implemented for reverting changes, we could think to have something that allow the user to apply all the suggestions for the entire file list.
It makes sense to you?

src/components/FullDetailView/HeaderSection.js
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
```
jeffmcaffer: This and the other new files are not inherently part of the full details view. they are used there but they may be used elsewhere.Is there anything stopping them from being moved up a level and added to top level index.js?

src/components/FullDetailView/HeaderSection.js
```diff
@@ -0,0 +1,53 @@
...
+import { Row, Button, Col } from 'react-bootstrap'
+import isEmpty from 'lodash/isEmpty'
+import { getBadgeUrl } from '../../api/clearlyDefined'
+import Definition from '../../utils/definition'
+import ButtonWithTooltip from '../Renderers/ButtonWithTooltip'
+
+const HeaderSection = props => {
+  const { definition, modalView, changes, renderContributeButton, handleClose, handleSave, handleRevert } = props
+  const { item } = definition
+  const scores = Definition.computeScores(item)
```
jeffmcaffer: is it still called `computeScores`? By now the service computes these and the values come back from the service in the definitions

src/components/FullDetailView/InnerDataSection.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+import React from 'react'
+import PlaceholderRenderer from '../Renderers/PlaceholderRenderer'
+import MonacoEditorWrapper from '../MonacoEditorWrapper'
+
+const InnerDataSection = ({ value, name, type = 'json', actionButton = null }) => {
```
jeffmcaffer: Why did you shift to the non-class syntax for these components (here and elsewhere). The rest of the code uses classes. Consistency wins!

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -196,6 +200,17 @@ export class FullDetailPage extends Component {
     })
   }
 
+  handleSuggestions() {
+    const key = `open${Date.now()}`
+    notification.open({
+      description:
+        'Another version of this defition has some recently updated data. \n \n Take a look at them and decide if to keep each data or discard it.',
```
jeffmcaffer: typo. definition

src/api/clearlyDefined.js
```diff
@@ -47,8 +48,42 @@ export function harvest(token, spec) {
...
+    url(`${CURATIONS}/${entity.toPath()}`, {
+      expanded: expandedPrs ? 'prs' : null,
+      state: pendingPrs ? 'pending' : null
+    }),
+    token
+  )
+}
+
+/**
+ * List all of the curations (if any) using the given coordinates as a pattern to match
```
jeffmcaffer: I feel like this code is duplicated from some other PR or something. The use of boolean params should be discouraged. Use an array of enum values in an `expand` arg or some such. here and above. More extensible and semantically valuable.

src/actions/ui.js
```diff
@@ -76,6 +80,19 @@ export function uiContributionGetData(token, entity) {
   return getPrDataAction(token, entity)
 }
 
+/**
+ * Get the curation in the given PR relative to the specified coordinates
+ * @param  {} token
```
jeffmcaffer: I'm sure that I added feedback on comments like these somewhere but don't see them. The empty params are a liability. Here for example, the comment param list does not match the actual param list. The comment params don't add any value. Unless the naming of things is deficient, the code should be self describing. 

The challenges with adding comments 
* easy to get out of date
* imply that things are more complicated than they otherwise seem from the code naming etc

Comments are great if their value outweighs the above issues. If we were doing APIs and publishing API doc then it would IMHO be a different situation. Even then, ideal is that the method and arg naming is semantically useful.

src/components/FullDetailView/HeaderSection.js
storrisi: If you think that HeaderSection and the other render components could be reused somewhere else, we could definitely move them as reusable components.

src/components/FullDetailView/HeaderSection.js
storrisi: I've to check this, this was an implementation of yours, related to show the Score image

src/api/clearlyDefined.js
storrisi: Part of this PR is used in #320 , due to the refactor of the FullDetailsView components.
I suggest to consider that one, and than update this one when the other one is merged, so we wouldn't have any duplicated implementations.

src/components/FullDetailView/HeaderSection.js
jeffmcaffer: I'd flip this around and say "if you think they **can't** be used anywhere else, leave them here. Otherwise, move them to be reusable"

jeffmcaffer:
@storrisi, I am deferring more comments here until #320 is merged. the duplicate code is causing too much confusion. Once we have merged the other one, please let me know when this one is good to go and I will re-review.

storrisi:
@jeffmcaffer i've updated the PR with the implementations merged by #320 .
Did you had the chance to check my comments above in this PR about the Suggestions View, and some other questions?

storrisi:
@jeffmcaffer i saw that you merged this, but there was some outstanding questions related on how to show the suggestions to the user, as you can see [here](https://github.com/clearlydefined/website/pull/316#issuecomment-426307402) 

Should we had to point these out in a new issue?

jeffmcaffer:
That would be great. Thanks

PR #323, opened 2018-10-05
Issue #257 Environments configuration

### LINKED ISSUE #257: Add .env file to manage development environments
As mentioned in [this discussion](https://github.com/clearlydefined/website/pull/248#discussion_r211091451), as `create-react-app` expects, development enviroment should be managed by a .env config file, in order to keep `src/api/clearlyDefined.js` while switching the enviroments.

storrisi:
This would introduce a standard environment configuration for the developer experience and deployment.
As documented [here](https://gist.github.com/storrisi/57cc66324c40a8987f8275c77c96f30d) this implementation changes the way of how the development environment are addressed into the app, using the standard approach of `.env` files.

The developer has to configure his local setup in order to use the proper configuration.

This also removes any hardcoded information from the code and fixes #257 

The documentation should be updated and i'm gonna open a specific issue on the doc repo.


examples/sample.env
```diff
@@ -0,0 +1 @@
+REACT_APP_API_URL='https://dev-api.clearlydefined.io'
```
jeffmcaffer: We should have preconfigured .env files for local, dev and prod. Forcing a user know about and create these files is error prone and adds friction. For this scenario (website) we do not have any secrets so its quite reasonable to put the files in the repo.

side note, in effect we have just moved the code that was in `clearlydefined.js` out into the config and build steps. The new approach is more extensible etc so good.

src/api/clearlyDefined.js
```diff
@@ -4,29 +4,7 @@
...
-      return API_DEVELOP
-    case 'production':
-      // TODO this needs to be replaced when we do a prod deployment. We want a "production" build deployed in
-      // the dev environment but of course it will need to point to the dev server. Don't know how to do that.
-      return API_DEVELOP
-    default:
-      return API_LOCAL
-  }
-}
+export const apiHome = process.env.REACT_APP_SERVER || process.env.REACT_APP_API_URL
```
jeffmcaffer: why have two different env var possibilities? Can we just use one or the other?

package.json
```diff
@@ -53,8 +54,10 @@
     "build-css": "node-sass src/ -o src/",
     "watch-css": "npm run build-css && node-sass src/ -o src/ --watch --recursive",
     "start-js": "react-scripts start",
-    "start": "npm-run-all -p build watch-css start-js",
-    "build": "npm run build-css && npm run copy-monaco && react-scripts build",
+    "start": "env-cmd --fallback ./.env.development npm-run-all -p build watch-css start-js",
+    "start:local": "env-cmd --fallback ./.env.local npm-run-all -p build watch-css start-js",
+    "build": "env-cmd --fallback ./.env.production npm run build-css && npm run copy-monaco && react-scripts build",
+    "build:dev": "env-cmd --fallback ./.env.development npm run build-css && npm run copy-monaco && react-scripts build",
```
jeffmcaffer: there is likely a bunch I am missing here. where/how/when is `build:dev` used?

I suspect that most of my questions will be resolved by some examples as mentioned above.

examples/sample.env
AlexWebYourmind: @jeffmcaffer I ran a few days ago with @storrisi  through this. While right now we don't have any Secrets, we should still keep the configuration outside the repo. As the application grows we might end up needing to add new configurations (some of which env specific) and starting with the right foot seems like a good idea (also you get security warnings as you commit .env files, and IMHO we should lead by example and stick to security standards). 

As you can see, we still have added samples + documentation to help developers in understanding the correct usage.

package.json
```diff
@@ -53,8 +54,10 @@
     "build-css": "node-sass src/ -o src/",
     "watch-css": "npm run build-css && node-sass src/ -o src/ --watch --recursive",
     "start-js": "react-scripts start",
-    "start": "npm-run-all -p build watch-css start-js",
-    "build": "npm run build-css && npm run copy-monaco && react-scripts build",
+    "start": "env-cmd --fallback ./.env.development npm-run-all -p build watch-css start-js",
```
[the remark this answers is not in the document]
AlexWebYourmind: @storrisi  I see what you have done here. The idea of using --fallback so that when running `npm start`  the configuration will be read from the **.env** default file if no  specific **.env.<environment>** file is present. Good thinking, but I think that if the user needs to run locally with different configurations other than "development" or "default" we might end up with a few "complications".

@jeffmcaffer watch-css is used when starting a local env, but when deploying on environments I do expect that just the **build** command will be executed or an env specific instruction, if we follow the **, build:env** approach suggested by @storrisi .


Anyways, I would rather think about this implementation from a different angle, and provide a way for both a user and a CI/CD process to provide any .env configuration without having to launch a different **npm script** at each time.

I think that ideally, we could proceed with one of the following 2 approaches:
###  Use only a .env file when running builds 
we could have only one build/start script reading configurations only and exclusively from the default **.env** file. A developer can then run a specific build as follows:

- Copy the **sample.env** configuration into a local **.env** file
- Change the configuration according to the desired one (Production APIs, Development etc)
- Run **npm start** which will build and serve the app reading from the **.env** file
Similarly, our **CI/CD** process can generate the **.env** file, by reading it from an **environment specific** file or directly from the CI SECRETS configuration, Copy over the given configuration, run build + deployment on any given envinroment.  Long story short, the CI/CD always injects the **development** or **production** config into a **.env** file, hence triggers the build

### Modify the NPM build & start scripts  so that external arguments are provided
We could alternatively change the npm scripts to pass a specific ".env.environment"  file externally
The difference in the previous approach would consist in the developer and the CI  launching**npm start** and **build** scripts providing the name of the specific **.env.development ** or **.env.anything** files to be used.

- ### use 1 .rc file
see https://www.npmjs.com/package/env-cmd#rc-file-usage

Given the --fallback instruction, the result would be a merge  
 

 


**NOTE:** While we can commit samples/ public configurations on the repository, so developers know how to setup their **.env** file (they can just copy-paste and fill in the specific conf), we should still keep the **.env** or **.env.environment** files outside the repo. This would give us the flexibility to keep the same approach when introducing new configurations like for example API_ACCESS_KEYS, and similar.


 @storrisi @jeffmcaffer thoughts?

examples/sample.env
jeffmcaffer: I'm certainly sympathetic to security concerns but I'll argue that as a design principle we explicitly want to NOT have secrets in the web site. It is just a generated static site so secrets would then be embedded in static downloadable content. If we do have secrets then we will have had some major architecture shifts and how to populate secrets will be part of that. We should optimize for simplicity now and add complexity as/if needed. 

Having to add more documentation to help users understand is another red flag.

If having .env files checked in causes security warnings (like what and from what system), then we should do something different or turn them off.

package.json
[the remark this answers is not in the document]
storrisi: Actually, an .env file is requested by default.
env-cmd allows only to set different environment files, which are consumed by the application or by the other instructions

src/api/clearlyDefined.js
storrisi: I've checked the history of the file and it appears that REACT_APP_SERVER is there from the first day.
Maybe this implementation is coming from something like a boilerplate?
I think we could remove it since REACT_APP_API_URL should be always available

jeffmcaffer: Yes, we have always called it APP_SERVER. This PR introduces a new var. I'm not fussed which one gets used but do want there to be just one.

jeffmcaffer:
This is what I am thinking the developer experience should be like

* `npm start` builds and runs the site pointing it at either localhost or dev apis (we can pick a default)
* `npm start:local-api` builds and runs the site pointing it at localhost for the API
* `npm start:dev-api` builds and runs the site pointing it at dev-api.clearlydefined.io for the API
* `npm start:prod-api` builds and runs the site pointing it at api.clearlydefined.io for the API

These should all just work out of the box, no fussing about copying files and configuring urls or ...

On the server we have a docker file that runs and the REACT_APP_SERVER env var is set in the configuration of the docker build so nothing is really needed there. Now if we wanted to source control the dev and prod environments then we shoudl talk about putting those settings files and configuring `build` scripts

.env.local
```diff
@@ -0,0 +1 @@
+REACT_APP_SERVER='https://localhost:4000'
```
jeffmcaffer: the use of .local here may be confusing. `.local` has specific meaning in this space. (seems to mean "my local thing that overrides the values as specified in the thing named my name but without .local") Rather, I suggest using `env.localhost` as the filename.

storrisi: I've update the name of the file as you suggested, and the related script as well

jeffmcaffer: do we still need this file?

PR #335, opened 2018-10-17
Issue #183 - DnD URLs from all supported providers

### LINKED ISSUE #183: DnD URLs from all supported providers
A user browsing GitHub, npmjs, Maven, ... should be able to take common URLs form the provider and drop them on ClearlyDefined such that the corresponding component is added to the list.

storrisi:
Fixes #183 
screen recording 2018-10-17 at 08 24 pm

### DIFF
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -4,7 +4,6 @@
 import React, { Fragment } from 'react'
 import { connect } from 'react-redux'
 import { Row, Col, Button, DropdownButton, MenuItem, OverlayTrigger, Tooltip } from 'react-bootstrap'
-import Dropzone from 'react-dropzone'
 import pako from 'pako'
 import throat from 'throat'
 import base64js from 'base64-js'
@@ -257,11 +256,40 @@ class PageDefinitions extends AbstractPageDefinitions {
     this.props.dispatch(uiNotificationNew({ type: 'info', message, timeout: 5000 }))
   }
 
-  onDrop(acceptedFiles, rejectedFiles) {
+  onDragOver = e => e.preventDefault()
+  onDragEnter = e => e.preventDefault()
+
+  onDrop = e => {
+    e.preventDefault()
+    const text = e.dataTransfer.getData('Text')
+
+    if (text) {
+      this.onTextDrop(text)
+    } else {
+      const files = Object.values(e.dataTransfer.files)
+      const checkedFiles = EntitySpec.checkDroppedFiles(files)
+      const { acceptedFiles, rejectedFiles } = checkedFiles
+
+      if (acceptedFiles.length) this.onFileDrop(acceptedFiles)
+      if (rejectedFiles.length) this.onDropRejected(rejectedFiles)
+    }
+  }
+
+  onTextDrop = url => {
     const { dispatch } = this.props
-    if (!acceptedFiles.length) return
+    const path = EntitySpec.fromUrl(url)
+
+    if (path.errors) dispatch(uiNotificationNew({ type: 'warning', message: path.errors, timeout: 5000 }))
+    else this.onAddComponent(path)
+  }
+
+  onFileDrop(files) {
+    const { dispatch } = this.props
+
+    if (!files.length) return
+
     dispatch(uiNotificationNew({ type: 'info', message: 'Loading component list from file(s)', timeout: 5000 }))
-    acceptedFiles.forEach(file => {
+    files.forEach(file => {
       const reader = new FileReader()
       reader.onload = () => {
         const listSpec = this.loadListSpec(reader.result, file)
@@ -292,15 +320,14 @@ class PageDefinitions extends AbstractPageDefinitions {
 
   dropZone(child) {
     return (
-      <Dropzone
-        accept="application/json"
-        disableClick
+      <div
+        onDragOver={this.onDragOver}
+        onDragEnter={this.onDragEnter}
         onDrop={this.onDrop}
-        onDropRejected={this.onDropRejected}
         style={{ position: 'relative' }}
       >
         {child}
-      </Dropzone>
+      </div>
     )
   }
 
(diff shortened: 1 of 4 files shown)


src/utils/entitySpec.js
```diff
@@ -15,6 +15,24 @@ const toLowerCaseMap = {
...
 
+const NPM_WEBSITE = 'npmjs.com'
+const GITHUB_WEBSITE = 'github.com'
+const MAVEN_WEBSITE = 'mvnrepository.com'
+const NUGET_WEBSITE = 'nuget.org'
+const PYPI_WEBSITE = 'pypi.org'
+const RUBYGEM_WEBSITE = 'rubygems.org'
+
+const providerPath = {
+  [NPM_WEBSITE]: 'npm/npmjs/-',
```
dabutvin: The namespace is not always going to be `-` here

For example https://www.npmjs.com/package/@types/lodash it will be `@types`

src/utils/entitySpec.js
```diff
@@ -15,6 +15,24 @@ const toLowerCaseMap = {
...
+  [GITHUB_WEBSITE]: 'git/github',
+  [PYPI_WEBSITE]: 'pypi/pypi/-',
+  [MAVEN_WEBSITE]: 'maven/mavencentral',
+  [NUGET_WEBSITE]: 'nuget/nuget/-',
+  [RUBYGEM_WEBSITE]: 'nuget/nuget/-'
```
dabutvin: Copy paste bug Ruby != Nuget

src/utils/entitySpec.js
```diff
@@ -15,6 +15,26 @@ const toLowerCaseMap = {
   mavencentralsource: NONE
 }
 
+const NPM_WEBSITE = 'npmjs.com'
```
jeffmcaffer: Rather than a destructured set of fields to capture each provider's details, we should have a first class object to represent a provider. That would have the path, as well as the function that teases apart the dropped URL and forms coordinates. Then we can have provider support plugged in by just discovering all the provider objects. There is more work to do other places to make that a reality but lets start down that path here.

src/utils/entitySpec.js
```diff
@@ -15,6 +15,26 @@ const toLowerCaseMap = {
...
+const GITHUB_WEBSITE = 'github.com'
+const MAVEN_WEBSITE = 'mvnrepository.com'
+const NUGET_WEBSITE = 'nuget.org'
+const PYPI_WEBSITE = 'pypi.org'
+const RUBYGEM_WEBSITE = 'rubygems.org'
+
+const providerPath = {
```
jeffmcaffer: These are the defaults but not necessarily always true. For example, we should support npms coming from github repos (we actually use this in clearlydefined itself) and release zips coming from GitHub. So the provider function that teases apart the dropped url should be the one that forms the path.

src/utils/entitySpec.js
jeffmcaffer: Should also add npmjs.org (it maps to .com). This is another reason to have this as an object. We could then just plug the same object in twice and have it handle both cases.

src/components/PageDefinitions.js
```diff
@@ -257,11 +256,40 @@ class PageDefinitions extends AbstractPageDefinitions {
     this.props.dispatch(uiNotificationNew({ type: 'info', message, timeout: 5000 }))
   }
 
-  onDrop(acceptedFiles, rejectedFiles) {
+  onDragOver = e => e.preventDefault()
```
jeffmcaffer: Can we factor all of this out into a different component? For exapmle, we'll want to DnD onto the harvest page as well.

src/utils/entitySpec.js
storrisi: I think this could a be a new requirement. Can we address this in a new specific issue?

src/utils/entitySpec.js
storrisi: Just to be sure to have understood well the suggestion:
- a Provider class should be implemented
- a specific Provider (like NpmProvider, and so on) should be implemented based on the generic Provider object, exposing specific methods and values
- each provider should contain specific rules on how to recognize a URL (like .com or .org extension for npmjs)
- a function that parses the URL and recognize which provider object is related

Is that right?

src/utils/entitySpec.js
jeffmcaffer: Its not new, the system design has always been this way. That is why we have type and provider as separate concepts. We've not done a lot of instances of it but would like to avoid taking implementation approaches that would prevent that. I'm fine getting something working in a simple form. I think the suggestion above to have an object that captures the behavior of a particular provider and/or type would address this.

src/utils/entitySpec.js
jeffmcaffer: More or less. Just imagine that a month from now someone wants to add support for some new provider. Do they roam about the code adding a bit here, a bit there, tweak and if or switch statement, ... of do they implement a new class with a bit of data and a handful of methods? I'm hoping for the latter.

storrisi: The idea i have is to build a generic Provider class, with shared structure and methods.
A specific provider could be an extension of the generic one, this would require to the developer to only build specific methods

jeffmcaffer: sure. lets not over engineer but at least some structure here would be good

storrisi: I will provide a refactor of this code while working on #168 , which actually adds a new type of supported URL

storrisi: @jeffmcaffer would you take a look at https://github.com/WebYourMind/website/pull/51/files#diff-b83a5a3eca3acc3a9eebddb8273448f3

I've implemented a solution proposal to address the Provider structure.

As i mentioned before, the idea is to allow the user to implement new specific providers, which are actually an extension of a generic one.

The Provider class should contain all the methods usable on each single specific provider.

Issue #24, opened 2018-01-21
testing strategy for web ui


jeffmcaffer:
Need to investigate the right/best way to test React components and apps, reducers, ...

* Pick a strategy
* implement example tests for each different kind of thing (component, container, reducer, ...)
* hook into build to ensure tests are run before deployment

mynameistechno:
For reference https://facebook.github.io/jest/docs/en/tutorial-react.html

mynameistechno:
For end-to-end testing we use nightwatchjs, but this still leverages selenium. Recently Chrome supports headless mode, and there are some newer browser automation tooling built around this that we should explore if we want to have browser-level/end-to-end type tests.

jeffmcaffer:
Great. I am new to this space so any help here would be greatly appreciated. Would be awesome if someone picked and did some examples.

AlexWebYourmind:
@mynameistechno  nightwatchjs (leveraging selenium) is a  great option indeed for E2E. 

As an alternative, may I also recommend to checkout cucumber (https://cucumber.io/) which has its own js implementation --> https://github.com/cucumber/cucumber-js and lets you write plain english tests (after creating different "grammars") and helps in adopting a BDD/TDD driven approach (but that's not a constraint).

Cucumber can of course leverage Selenium and I have been using that on many projects and always turned out quite useful. Also tests can be written resembling User Stories and Acceptance Criteria where needed ( I know we are not adopting SCRUM yet).

Long story short, we could start working towards introducing E2E Tests covering all the CD Use Cases (from a User/Actor perspective) and gradually achieve a good test coverage.

This is definitely doable with Nightwatch as well, it's just a matter of deciding which E2E tool we prefer.

mynameistechno:
> @mynameistechno nightwatchjs (leveraging selenium) is a great option indeed for E2E.

Yes I was thinking the same. I.e. given the current lack of unit tests, writing some end-to-end tests for the happy paths would be the easiest way to add some sort of coverage to help prevent regressions. E2E tests shouldn't be used for code coverage however, since e2e tests don't have visibility into the code and are abstracted away from them. Having said that, they provide great functional coverage.

> As an alternative, may I also recommend to checkout cucumber (https://cucumber.io/) which has its own js implementation --> https://github.com/cucumber/cucumber-js

[...]

PR #336, opened 2018-10-18
Tests Implementation

storrisi:
Related to issue #24 

* Updated react-scripts to 2.0.4

* Added some samples, and actually working, tests implementation of components, utils, actions, async actions using fetch-mock, reducer tests and connected components

* Proposal of scaffolding method for __tests__ folders

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,351 +4,1048 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@babel/code-frame": {
-      "version": "7.0.0-beta.51",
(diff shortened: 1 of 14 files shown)


jeffmcaffer:
Nice! Assuming the react-scripts change works on windows, mac and linux then we should put this in. I did a quick look at the tests and they seem to make sense. Never done react ui tests so will have to write a few myself to get the full idea.

Note the merge conflict on package-lock.json

storrisi:
Thank @jeffmcaffer , that's quite strange, because when i've opened the PR there wasn't any conflict.
Maybe because of some recent merge. I'm going to fix it.

storrisi:
@jeffmcaffer i've fixed the conflicts. Thanks

PR #345, opened 2018-10-26
Add or Change versions of a definition

### LINKED ISSUE #160: Add more versions
Once a user has found one version of a component it is often interesting to look at others. The component list should allow users to add additional versions by clicking a convenient button/link on a list entry and then multi-selecting from a list of known versions. These new versions should be added to the list directly after the entry in which a user clicks (subject to sorting and filtering settings).

### LINKED ISSUE #185: Allow a different version to be selected
The user may have selected or DnD'd the wrong version of a component. Allow them the select a different version. This changes the selected list entry to be an entry for the new version.

If the entry has edits, the user if prompted to confirm abandoning the edits. The edits are NOT carried forward.

storrisi:
Fixes #160 , Fixes #185 

screen recording 2018-10-26 at 05 14 pm
screen recording 2018-10-26 at 05 15 pm

Some refactoring of the ComponentsList has been done, putting in a separate Component the buttons renderer.
The implementation is also provided with tests.

### DIFF
--- a/src/api/clearlyDefined.js
+++ b/src/api/clearlyDefined.js
@@ -16,6 +16,7 @@ export const ORIGINS_NUGET = 'origins/nuget'
 export const ORIGINS_MAVEN = 'origins/maven'
 export const ORIGINS_PYPI = 'origins/pypi'
 export const ORIGINS_RUBYGEMS = 'origins/rubygems'
+export const ORIGINS = namespace => `origins/${namespace}`
 
 export function getHarvestResults(token, entity) {
   // TODO ensure that the entity has data all the way down to the revision (and no more)
@@ -159,6 +160,10 @@ export function getNugetSearch(token, path) {
 
 export function getNugetRevisions(token, path) {
   return get(url(`${ORIGINS_NUGET}/${path}/revisions`), token)
+}
+
+export function getRevisions(token, path, namespace) {
+  return get(url(`${ORIGINS(namespace)}/${path}/revisions`), token)
 }
 
 // ========================== utilities ====================
--- a/src/components/AbstractPageDefinitions.js
+++ b/src/components/AbstractPageDefinitions.js
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-import React, { Component } from 'react'
+import React, { Component, Fragment } from 'react'
 import { Modal, Grid, DropdownButton, MenuItem, FormGroup, InputGroup, FormControl, Button } from 'react-bootstrap'
 import compact from 'lodash/compact'
 import filter from 'lodash/filter'
@@ -9,6 +9,8 @@ import find from 'lodash/find'
 import get from 'lodash/get'
 import set from 'lodash/set'
 import sortBy from 'lodash/sortBy'
+import notification from 'antd/lib/notification'
+import AntdButton from 'antd/lib/button'
 import { curateAction } from '../actions/curationActions'
 import { login } from '../actions/sessionActions'
 import { ComponentList, Section, ContributePrompt } from './'
@@ -17,6 +19,7 @@ import { uiBrowseUpdateFilterList } from '../actions/ui'
 import EntitySpec from '../utils/entitySpec'
 import Definition from '../utils/definition'
 import Auth from '../utils/auth'
+import VersionSelector from './Navigation/Ui/VersionSelector'
 
 const sorts = [
   { value: 'license', label: 'License' },
@@ -77,6 +80,9 @@ export default class AbstractPageDefinitions extends Component {
     this.onRemoveAll = this.onRemoveAll.bind(this)
     this.collapseAll = this.collapseAll.bind(this)
     this.renderSavePopup = this.renderSavePopup.bind(this)
+    this.renderVersionSelectopPopup = this.renderVersionSelectopPopup.bind(this)
+    this.showVersionSelectorPopup = this.showVersionSelectorPopup.bind(this)
+    this.applySelectedVersions = this.applySelectedVersions.bind(this)
     this.contributeModal = React.createRef()
   }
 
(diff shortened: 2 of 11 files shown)


src/components/AbstractPageDefinitions.js
```diff
@@ -436,6 +442,70 @@ export default class AbstractPageDefinitions extends Component {
...
+          const key = `open${Date.now()}`
+          const NotificationButtons = (
+            <Fragment>
+              <AntdButton
+                type="primary"
+                size="small"
+                onClick={async () => {
+                  await this.onRemoveComponent(selectedComponent)
+                  await this.onAddComponent(
+                    EntitySpec.fromCoordinates({ ...selectedComponent, revision: versions, changes: {} })
```
jeffmcaffer: is it really `versions` (with an s)? how does that work?

src/components/AbstractPageDefinitions.js
```diff
@@ -436,6 +442,70 @@ export default class AbstractPageDefinitions extends Component {
...
+    this.setState({ showVersionSelectorPopup: true, multipleVersionSelection: multiple, selectedComponent: component })
+  }
+
+  applySelectedVersions(versions) {
+    const { multipleVersionSelection, selectedComponent } = this.state
+    if (!multipleVersionSelection) {
+      return this.setState({ showVersionSelectorPopup: false }, async () => {
+        if (selectedComponent.changes) {
+          const key = `open${Date.now()}`
+          const NotificationButtons = (
```
jeffmcaffer: this should be factored out as a separate function that can be reused. Also simplifies this code making it easier to understand

src/components/AbstractPageDefinitions.js
```diff
@@ -436,6 +442,70 @@ export default class AbstractPageDefinitions extends Component {
...
+                size="small"
+                onClick={async () => {
+                  await this.onRemoveComponent(selectedComponent)
+                  await this.onAddComponent(
+                    EntitySpec.fromCoordinates({ ...selectedComponent, revision: versions, changes: {} })
+                  )
+                  notification.close(key)
+                }}
+              >
+                Confirm
```
jeffmcaffer: Stick with OK and Cancel rather than Confirm/Dismiss

src/components/AbstractPageDefinitions.js
```diff
@@ -436,6 +442,70 @@ export default class AbstractPageDefinitions extends Component {
...
+              </AntdButton>{' '}
+              <AntdButton type="secondary" size="small" onClick={() => notification.close(key)}>
+                Dismiss
+              </AntdButton>
+            </Fragment>
+          )
+          notification.open({
+            message:
+              'This definition has some changes. Do you confirm to change the version? All the unsaved changes will be lost.',
```
jeffmcaffer: Simplify to `This definition has some changes. All the unsaved changes will be lost.`

src/components/Navigation/Ui/VersionSelector.js
```diff
@@ -0,0 +1,110 @@
...
+    this.setState(
+      {
+        options: [],
+        selected: [],
+        label: ''
+      },
+      () => onSave(selected)
+    )
+  }
+  onClose = () => {
```
jeffmcaffer: Consider formatting on fewer lines like this. (here and elsewhere)
```javascript
 onClose = () => {
    const { onClose } = this.props
    this.setState({ options: [], selected: [], label: '' }, () => onClose())
  }
```

src/components/Navigation/Ui/VersionSelector.js
```diff
@@ -0,0 +1,110 @@
...
+const Option = Select.Option
+class VersionSelector extends Component {
+  static propTypes = {
+    multiple: PropTypes.bool,
+    show: PropTypes.bool,
+    onClose: PropTypes.func,
+    onSave: PropTypes.func,
+    component: PropTypes.object
+  }
+  constructor(props) {
```
jeffmcaffer: blank line between functions (here and elsewhere(

src/components/Navigation/Ui/VersionSelector.js
```diff
@@ -0,0 +1,110 @@
...
+    return (
+      <Modal show={show} onHide={this.onClose}>
+        <Modal.Header closeButton>
+          <Modal.Title>{label}</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <FormGroup>
+            <Select
+              mode={multiple && 'multiple'}
+              style={{ width: '100%' }}
```
jeffmcaffer: bit bummed by how wide the modal is in the UI. Could trim that down substantially

src/utils/entitySpec.js
```diff
@@ -180,4 +180,23 @@ export default class EntitySpec {
...
+        ? revision.sha
+        : `${revision.tag} (${revision.sha})`
+      : revision
+  }
+
+  static getEntityName(component) {
+    return component.namespace ? `${component.namespace}/${component.name}` : component.name
+  }
+  static getEntityType(component) {
+    return component.provider === 'github' || component.provider === 'rubygems' ? component.provider : component.type
```
jeffmcaffer: don't understand this. type and provider are separate concepts.

src/utils/entitySpec.js
```diff
@@ -180,4 +180,23 @@ export default class EntitySpec {
   toString() {
     return this.toPath()
   }
+
+  static getRevisionToKey(revision, definition) {
+    return definition.provider === 'github' ? revision.sha : revision
```
jeffmcaffer: `revision` should always be a string.  Feels like this method and `getRevisionToString` are helpers for some specific scenarios and are just being dumped on this class. Suggest they go somewhere else as they do not match the nature of EntitySpec itself.

src/components/Navigation/Ui/VersionSelector.js
```diff
@@ -0,0 +1,110 @@
...
+              {options.map(option => (
+                <Option key={EntitySpec.getRevisionToKey(option, component)}>
+                  {EntitySpec.getRevisionToString(option, component)}
+                </Option>
+              ))}
+            </Select>
+          </FormGroup>
+        </Modal.Body>
+        <Modal.Footer>
+          <div>
```
jeffmcaffer: as a point of interest, why the extra `<div>`?

jeffmcaffer:
Overall an interesting approach. I like the ability to add multiple versions. Nice. Do you think we could accomplish the same/similar with a drop down similar to what you see in the Harvest experience? Feels like it would be a bit more "in-place" than having a modal pop up and go away. Fewer clicks, visually smoother.

src/utils/entitySpec.js
storrisi: I saw a lot of methods in the clearlyDefined.js that returns quite the same string.
So what i've done is implement a constant that should help to build the revisions path, and not have a lot of duplicate code: `export const ORIGINS = namespace => `origins/${namespace}` , what i've done is to find a way to build that string in the proper way.
For example, it should be 'origins/github' or 'origins/npm'

This is a use case:
- if i want to get all the versions of `npm/npmjs/-/async/2.6.0` , i should pass to the getRevisions function the type (npm) and the name (async) to return `origins/npm`
- if i want to do the same thing for `git/github/github/fetch/v2.0.3` , i should pass to the same function the provider (github) and the name (async), to return `origins/github`

Maybe the naming is wrong, but i hope the implementation can be clear and helpful

src/components/AbstractPageDefinitions.js
storrisi: This function is built to accept either a single version (if the user used the Switch functionality) or an array (if the user used the Add functionality).
So this params could contain a single value or an array, it depends on the `multipleVersionSelection` into the component state

src/components/AbstractPageDefinitions.js
storrisi: Actually there was already a reusable component. I missed to use it in this point, thanks for catching it

src/utils/entitySpec.js
storrisi: I have renamed the function in `getEntityOrigin`, maybe it could be more understandable the purpose?

storrisi:
@jeffmcaffer i've updated the PR based on your suggestions.

> Do you think we could accomplish the same/similar with a drop down similar to what you see in the Harvest experience?

If you mean an inline dropdown in the list, we should think something that could fit there.
But in terms of clicks, i think the user will have exactly the same steps (choose to Switch or Add, select the version, confirm or undo)

src/utils/entitySpec.js
storrisi: I've moved these methods into the Definition class.

src/components/AbstractPageDefinitions.js
jeffmcaffer: ok but EntitySpecs are not intended/expected to have multiple entries in the revision property.

src/api/clearlyDefined.js
```diff
@@ -16,6 +16,14 @@ export const ORIGINS_NUGET = 'origins/nuget'
 export const ORIGINS_MAVEN = 'origins/maven'
 export const ORIGINS_PYPI = 'origins/pypi'
 export const ORIGINS_RUBYGEMS = 'origins/rubygems'
+export const ORIGINS = {
```
jeffmcaffer: This feels leaky. Users of the clearlydefined api should just indicate what they want and not have to pass in elements of the url to be accessed. So in `getRevisions` it would be better for users to pass in what they have and we figure out the right URL to use in the request.

src/utils/entitySpec.js
```diff
@@ -180,4 +182,11 @@ export default class EntitySpec {
   toString() {
     return this.toPath()
   }
+
+  static getEntityName(component) {
+    return component.namespace ? `${component.namespace}/${component.name}` : component.name
+  }
+  static getEntityOrigin(component) {
```
jeffmcaffer: see comment at the beginning. Feels leaky to expose something about the api URL. Especially here on a relatively generic data structure class.

src/api/clearlyDefined.js
storrisi: Right, i've updated the method avoiding the user to pass in directly URL informations.

src/components/AbstractPageDefinitions.js
storrisi: Yes, it never happens.
In case of multiple versions selected, there's a loop that iterates all the versions, and then calls the specific function to retrieve it:

```
versions.map(version => {
        this.onAddComponent(EntitySpec.fromCoordinates({ ...selectedComponent, revision: version }))
      })
```

PR #350, opened 2018-10-29
Issue #214 - harvesting form shows wrong tags when deleting request

### LINKED ISSUE #214: harvesting form shows wrong tags when deleting request
How to replicate:

1 Go to https://dev.clearlydefined.io/harvest
1  Add a git repo, select a tag.
1  Add another git repo, select a tag.
1  Delete first (top) repo.

What happens:

Tags for first repo stick around in pulldown, next to second repo's name

What should happen:

Second repo should have second repo's tags.

storrisi:
Fixes #214

### DIFF
--- a/src/components/HarvestQueueList.js
+++ b/src/components/HarvestQueueList.js
@@ -53,7 +53,8 @@ class HarvestQueueList extends React.Component {
   commitChanged(request, value) {
     const newRequest = clone(request)
     newRequest.revision = value ? value.sha : null
-    this.setState({ contentSeq: this.state.contentSeq + 1 })
+    newRequest.commit = value
+    this.setState({ ...this.state, contentSeq: this.state.contentSeq + 1 })
     this.props.onChange(request, newRequest)
   }
 
@@ -159,6 +160,7 @@ class HarvestQueueList extends React.Component {
   render() {
     const { loadMoreRows, listHeight, noRowsRenderer, list } = this.props
     const { sortOrder, contentSeq } = this.state
+
     return (
       <RowEntityList
         list={list}
@@ -174,7 +176,7 @@ class HarvestQueueList extends React.Component {
   }
 }
 
-function mapStateToProps(state, ownProps) {
+function mapStateToProps(state) {
   return {
     token: state.session.token
   }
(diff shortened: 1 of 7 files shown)


src/components/HarvestQueueList.js
```diff
@@ -53,7 +53,8 @@ class HarvestQueueList extends React.Component {
   commitChanged(request, value) {
     const newRequest = clone(request)
     newRequest.revision = value ? value.sha : null
-    this.setState({ contentSeq: this.state.contentSeq + 1 })
+    newRequest.commit = value
```
jeffmcaffer: feels odd to be adding a new property just for this case. Can we just be using `revision`?

src/components/HarvestQueueList.js
```diff
@@ -159,6 +160,7 @@ class HarvestQueueList extends React.Component {
   render() {
     const { loadMoreRows, listHeight, noRowsRenderer, list } = this.props
     const { sortOrder, contentSeq } = this.state
+
```
jeffmcaffer: remove the extra blanks here and elsewhere

src/components/MavenVersionPicker.js
```diff
@@ -14,7 +14,7 @@ export default class MavenVersionPicker extends Component {
 
   constructor(props) {
     super(props)
-    this.state = { customValues: [], options: [] }
+    this.state = { customValues: [], options: [], selected: props.request.revision ? [props.request.revision] : [] }
```
jeffmcaffer: All of the changes here and in the other pickers are duplicate. Need some way to either eliminate the need for the code or share it.

Related, it feels like this scenario is not that unusual and ought not be this hard to implement. This approach feels sort of subtle. I don't know a better one right now but if feels overly complicated. (could just be my naivety)

src/components/MavenVersionPicker.js
```diff
@@ -23,6 +23,13 @@ export default class MavenVersionPicker extends Component {
     this.getOptions('')
   }
 
+  componentWillReceiveProps(nextProps) {
+    this.setState(prevState => ({
```
jeffmcaffer: can this be

```javascript
this.setState({...this.state, selected: nextProps.request.revision ? [nextProps.request.revision] : []})
```

That is, is the function needed

src/components/HarvestQueueList.js
storrisi: The `commit` property is needed for the GitHubCommitPicker, so we had to add it in order to use it

storrisi:
@jeffmcaffer i've updated the PR based on your suggestions

PR #351, opened 2018-10-29
Issue #186 - DnD entry from on CD list to another

### LINKED ISSUE #186: DnD ClearlyDefined entry from one CD list to another
A user with two ClearlyDefined lists (e.g., two open tabs) can DnD an entry from one to the other.

Pending decisions:

* what happens to any pending edits. Ideally they would carry over. 
* is the origin entry removed? That is, is this a Move or a Copy operation. Suggest that at least initially it is a Copy as the user can easily delete the origin (red X).

storrisi:
Fixes #186 

screen recording 2018-10-29 at 07 58 pm


src/utils/utils.js
```diff
@@ -18,4 +18,20 @@ function difference(object, base) {
   return newValue
 }
 
-export { setIfValue, difference }
+function isJson(item) {
```
jeffmcaffer: Suggest the following. if `item` is an object, then it it is "json". if it is a string and parses then it too is "json". 

```javascript
function isJson(item) {
  if (!item) return false
  if (typeof item !== 'string') return true
  try {
    return !!JSON.parse(item)
  } catch (e) {
    return false
  }
}
```

src/components/TwoLineEntry.js
```diff
@@ -22,9 +22,27 @@ export default class TwoLineEntry extends React.Component {
...
+      isEmpty,
+      draggable,
+      item
+    } = this.props
     return (
-      <div className="two-line-entry">
+      <div
+        className="two-line-entry"
+        draggable={draggable}
+        onDragStart={ev => {
```
jeffmcaffer: Should the type be `application/json`?

Also, compact formatting.

```javascript
        onDragStart={event => event.dataTransfer.setData('application/json', JSON.stringify(item))}
```

src/components/PageDefinitions.js
```diff
@@ -279,12 +280,14 @@ export class PageDefinitions extends AbstractPageDefinitions {
...
   }
 
-  onTextDrop = url => {
+  onTextDrop = content => {
     const { dispatch } = this.props
-    const path = EntitySpec.fromUrl(url)
-
-    if (path.errors) dispatch(uiNotificationNew({ type: 'warning', message: path.errors, timeout: 5000 }))
-    else this.onAddComponent(path)
+    if (!isJson(content)) {
```
jeffmcaffer: if the drop had `type === 'application/json'` could also be dispatching based on that. may still be handy to allow dropping of arbitrary "text" that happens to be json...

PR #359, opened 2018-11-11
encasulate version selection origin handling

jeffmcaffer:
@storrisi this is what I meant about the management of origins and entity specs. The UI is trafficking in a plain object, EntitySpec should not know anything about origin URLs and the origin mapping, in the fullness of time, is a combo of type and provider (e.g., we will want to get `releases` from github and git things from gitlab, ...)


jeffmcaffer:
I'm going to proactively merge this to enable some of the other tweaks I'm doing. @storrisi, please do review and tweak further as needed.

PR #360, opened 2018-11-11
clean separation between drop and entityspec

jeffmcaffer:
There were some places where the DnD handling code got intermingled with EntitySpec.  This PR teases those apart and makes EntitySpec a bit more extensible. In the future we'll want the support for each entity origin etc to be plugged in (or at least all clustered together in one spot) rather than sprinkled across all the code as it is now. That will make it easier to add and manage different providers. This is a first small step in that direction.

### DIFF
--- a/src/utils/entitySpec.js
+++ b/src/utils/entitySpec.js
@@ -15,20 +15,46 @@ const toLowerCaseMap = {
   mavencentralsource: NONE
 }
 
-const NPM_WEBSITE = 'npmjs.com'
-const GITHUB_WEBSITE = 'github.com'
-const MAVEN_WEBSITE = 'mvnrepository.com'
-const NUGET_WEBSITE = 'nuget.org'
-const PYPI_WEBSITE = 'pypi.org'
-const RUBYGEM_WEBSITE = 'rubygems.org'
+const entityMapping = [
+  { hostnames: ['npmjs.com', 'npmjs.org'], parser: npmParser },
+  { hostnames: ['github.com'], parser: githubParser },
+  { hostnames: ['mvnrepository.com'], parser: mavenParser },
+  { hostnames: ['nuget.org'], parser: nugetParser },
+  { hostnames: ['pypi.org'], parser: pypiParser },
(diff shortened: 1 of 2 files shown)


src/utils/entitySpec.js
```diff
@@ -15,25 +15,47 @@ const toLowerCaseMap = {
...
+function npmParser(host, path) {
+  let namespace, name, version
+  const pathSegments = path.split('/')
+  if (pathSegments.length === 5) [, namespace, name, , version] = pathSegments
+  else [, name, , version] = pathSegments
+  return new EntitySpec('npm', 'npmjs', namespace, name, version)
 }
 
-const providerWebsiteValues = [NPM_WEBSITE, GITHUB_WEBSITE, MAVEN_WEBSITE, NUGET_WEBSITE, PYPI_WEBSITE, RUBYGEM_WEBSITE]
+function githubParser(host, path, error) {
```
storrisi: How `error` is used here?
I saw this in other methods, but it seems to be never used

jeffmcaffer: that was stale and is now removed

jeffmcaffer:
There are definite similarities. Some thoughts
* This is just a first step. Eventually we'll have a plugin mechanism that allows different parsers etc to be plugged in. Even more eventually we'll have complete support for a package manager plugged in (entity spec parsing, origin management, version selection, ...). At that point the parser methods etc will be fanned out and the lookup mechanism will consult the plugin registry. With this first step, the bits of code are distinct and separate.
* we do need a plug in mechanism to make this work. I've not done a lot of investigation on what's available/good. Suggestions more than welcome.
* Gists are a different thing. IMO they are better as a separate concept as the APIs etc should be more tuned to how they are used.
* EntitySpec is a simple data type. Previously it had various UI function (e.g., checkDroppedFiles) on it. This PR moves that into the UI code.
* I agree that the UI drop code needs to be factored out and reusable. I'd like to see all of the Drop related code in PageDefinitions.js moved out into some reusable Dropzone component.

PR #361, opened 2018-11-11
Add tests for entity spec

jeffmcaffer:
These tests are modeled after the ones from @storrisi  in #356 and will supersede those.


jeffmcaffer:
proactively merging

PR #366, opened 2018-11-12
Add support for saving/loading gist-based lists

jeffmcaffer:
Building on #356, this PR integrates gist support into the new provider and api approach. #356 is obsoleted by this (and a previously merged PR).

There is an open question:
* How to get a token that will allow the site to create a Gist in the user's account. The normal token we get does not have permissions AFAICT. Not sure how it worked in #356 as we use the same technique/token here and end up with a 404 on POST to create the Gist.

Probably should not merge until we sort this out.

### DIFF
--- a/src/actions/ui.js
+++ b/src/actions/ui.js
@@ -57,6 +57,19 @@ export function uiRedirect(to) {
 }
 
 let nextNotificationId = 0
+
+export function uiInfo(dispatch, message, timeout = 5000) {
+  dispatch(uiNotificationNew({ type: 'info', message, timeout }))
+}
+
+export function uiWarning(dispatch, message, timeout = 5000) {
+  dispatch(uiNotificationNew({ type: 'warning', message, timeout }))
+}
+
+export function uiDanger(dispatch, message, timeout = 5000) {
+  dispatch(uiNotificationNew({ type: 'danger', message, timeout }))
+}
+
 export function uiNotificationNew(message) {
   return { type: UI_NOTIFICATION_NEW, message: { ...message, id: nextNotificationId++ } }
 }
--- a/src/components/PageDefinitions.js
+++ b/src/components/PageDefinitions.js
@@ -13,7 +13,7 @@ import AntdButton from 'antd/lib/button'
 import chunk from 'lodash/chunk'
 import isEmpty from 'lodash/isEmpty'
 import { FilterBar } from './'
-import { uiNavigation, uiBrowseUpdateList, uiNotificationNew, uiRevertDefinition } from '../actions/ui'
+import { uiNavigation, uiBrowseUpdateList, uiRevertDefinition, uiInfo, uiWarning, uiDanger } from '../actions/ui'
 import { getDefinitionsAction } from '../actions/definitionActions'
 import { ROUTE_DEFINITIONS, ROUTE_SHARE } from '../utils/routingConstants'
 import EntitySpec from '../utils/entitySpec'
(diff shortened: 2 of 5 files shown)


src/components/PageDefinitions.js
```diff
@@ -275,12 +304,16 @@ export class PageDefinitions extends AbstractPageDefinitions {
...
 
   onTextDrop = content => {
-    const contentObject = asObject(content)
-    if (contentObject) return this.onAddComponent(EntitySpec.fromCoordinates(contentObject))
     try {
+      const contentObject = asObject(content)
+      // if the dropped text can be directly interpreted as JSON, assume it's a list
+      if (contentObject) return this.onAddComponent(EntitySpec.fromCoordinates(contentObject))
+      // if it's a URL to a gist, load the gist files
+      if (content.startsWith('https://gist.github.com')) return this.onGistDrop(content)
```
storrisi: I think it's not fair to put this kind of condition here.
We could end up managing different URLs (another one that we already know to manage is PR URL)
In my #356 this have been managed by an high level class Provider, that will address all the different providers, as you can see here https://github.com/clearlydefined/website/pull/356/files#diff-b83a5a3eca3acc3a9eebddb8273448f3R17

What do you think about it?

src/components/PageDefinitions.js
```diff
@@ -296,38 +329,38 @@ export class PageDefinitions extends AbstractPageDefinitions {
     )
   }
 
+  async onGistDrop(urlString) {
```
storrisi: As for the above comment, i would likely remove any kind of specific provider management from a single component.
These could be reusable methods for other pages or lists (like the one in the new Browse page) where we probably need to share them.

A utils class that manage all these different implementation it's the best way imo

src/actions/ui.js
```diff
@@ -57,6 +57,19 @@ export function uiRedirect(to) {
 }
 
 let nextNotificationId = 0
+
+export function uiInfo(dispatch, message, timeout = 5000) {
+  dispatch(uiNotificationNew({ type: 'info', message, timeout }))
+}
```
storrisi: In a redux-thunk context, this should usually become:
```
export function uiInfo(message, timeout = 5000) {
  return dispatch => {
  dispatch(uiNotificationNew({ type: 'info', message, timeout }))
  }
}
```

This would avoid to pass in the `dispatch` prop from the component to any action creator

src/components/PageDefinitions.js
```diff
@@ -39,7 +40,7 @@ export class PageDefinitions extends AbstractPageDefinitions {
         const definitionSpec = pako.inflate(base64js.toByteArray(path), { to: 'string' })
         this.loadFromListSpec(JSON.parse(definitionSpec))
       } catch (e) {
-        dispatch(uiNotificationNew({ type: 'warning', message: 'Loading components from URL failed', timeout: 5000 }))
+        uiWarning(dispatch, 'Loading components from URL failed')
```
storrisi: Using the above mentioned Redux method, the use of `dispatch` results not needed anymore

src/actions/ui.js
jeffmcaffer: These are just helper methods to make the code smaller. They are not actions per se. Perhaps I am missing the point. A complete working example would help.

src/components/PageDefinitions.js
jeffmcaffer: yes. the entire drop thing should be factored out and be pluggable to accept different handlers etc. The providers approach in #356 was too monolithic for this and mixed concerns.

src/components/PageDefinitions.js
jeffmcaffer: I'd like to have a drop component that can be configured with handlers. The current direction is towards that where the handlers just come in as an array and the drop handler searches for the appropriate handler and runs it. There will be future refinements to be sure.

PR #375, opened 2018-11-24
WIP : Issue #309 - Browse Page

### LINKED ISSUE #309: Interface for "components that need curation"
One way we can give folks a way to get going on curations is to provide a page on the website with random components that are missing declared licenses. We could then invite them to easily curate one of those components right there. It's hard to know where to start right now on which components need curations, so let's give folks a way in and pick out components we know need curations and display them to the user.

storrisi:
Fixes #309 

Initial WIP PR with some refactoring of the AbstractPageDefinitions / PageDefinitions, in order to clean the code, remove unused method, move shared methods into utils classes or HoC.

Most of the methods should be used by the new PageBrowse , so the refactor it's necessary to avoid code duplication.


src/components/Navigation/Pages/PageBrowse.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+function mapStateToProps(state) {
+  return {
+    token: state.session.token
+  }
+}
+
+function mapDispatchToProps(dispatch) {
+  return bindActionCreators(
```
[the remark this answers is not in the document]
storrisi: That's a Prettier default behaviour, i will check if we can avoid it

src/components/PageDefinitions.js
```diff
@@ -224,8 +148,49 @@ export class PageDefinitions extends AbstractPageDefinitions {
     )
   }
 
-  updateList(o) {
-    return uiBrowseUpdateList(o)
+  renderFilter(list, title, id) {
```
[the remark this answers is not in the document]
storrisi: I agree, i've moved only some blocks of code for now, this should become a separate component

src/utils/drop.js
```diff
@@ -0,0 +1,118 @@
...
+import trim from 'lodash/trim'
+import { ROUTE_CURATIONS } from '../utils/routingConstants'
+import { getGist } from '../api/github'
+import { uiInfo, uiWarning } from '../actions/ui'
+
+/**
+ * Abstract methods for Dropping functionality
+ *
+ */
+export default class Drop {
```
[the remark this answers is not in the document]
storrisi: I was starting building a utils class thinking of reusing all the related methods, but your comment pointed me out in the right direction.
We don't need a class of methods for the Drop functionality, but it should be a real Component, plugged in like the DropZone library

src/components/AbstractPageDefinitions.js
```diff
@@ -498,56 +321,223 @@ export default class AbstractPageDefinitions extends Component {
...
+    const path = component.toPath()
+    if (!component.revision) return uiWarning(dispatch, `${path} needs version information`)
+
+    !definitions.entries[path] &&
+      dispatch(getDefinitionsAction(token, [path])) &&
+      dispatch(getCurationAction(token, component))
+    dispatch(uiBrowseUpdateList({ add: component }))
+  }
+
+  dropZone(child) {
```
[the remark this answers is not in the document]
storrisi: Yes, for sure

storrisi:
Thanks @jeffmcaffer , the difference between user-managed and system-managed is interesting.
I will work in order to distinguish the two behaviour, probably creating some HoC or Abstract Components that contains the logic.
What do you think?

src/components/Navigation/Pages/PageBrowse.js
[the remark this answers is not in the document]
jeffmcaffer: If you just don;t put any line breaks in then Prettier will format it like this 

```javascript
function mapDispatchToProps(dispatch) {
  return bindActionCreators({ uiNavigation }, dispatch)
}
```

jeffmcaffer:
User vs system may be useful. I was just observing the difference as I read through your changes. Not sure if that is the major pivot or just a characteristic. From a code point of view if we use composition more then we don't have to think too hard about that as a hierarchy as things are a bit more mix and match. 

From a user experience point of view it may be useful to essentially capture user expectations (e.g., as a user I am constructing/managing a list vs observing an existing list).

storrisi:
@jeffmcaffer i've finalized the work here.
This is how the page looks like:
schermata 2018-11-30 alle 16 02 40

As you can see this page is quite a "mix" of a System Managed list , which is basically Read-Only, and a User Managed List.
This kind of page has all the System Managed List functionalities, but allow the user to edit the definitions and make a curation.
Beside that, the user can't modify the returned list, neither share it.

Two small advice:
- When a user selects a provider, the application calls a "/browse?pattern={provider}" API, which is not already implemented
- given that, i've hidden the "Browse" voice from the Navigation Menu. 
You can decide to wait to merge this PR until we have completed the service implementation, or merge it anyway and keep the page hidden.

src/actions/definitionActions.js
```diff
@@ -72,3 +82,22 @@ export function revertDefinitionAction(definition, values, name) {
     return dispatch(actions.success({ componentsWithoutChanges }))
   }
 }
+
+export function browseDefinitionsAction(token, entity, name) {
+  return dispatch => {
+    const actions = asyncActions(name)
+    dispatch(actions.start())
+    return getLowScoreDefinitions(token, entity).then(
```
jeffmcaffer: this is ok for now but in the end we'll want to pass a filter/sort set to guide the query. it is not just low score. We will eventually also have to do pages etc. Perhaps just generalize the naming and we can add the additional filtering/sorting capabilities later

jeffmcaffer: lets use async/await patterns. here and elsewhere

src/actions/definitionActions.js
```diff
@@ -72,3 +82,22 @@ export function revertDefinitionAction(definition, values, name) {
...
+      result => {
+        dispatch(actions.success({ add: result }))
+        const toAdd = map(result, component => EntitySpec.validateAndCreate(component.coordinates)).filter(e => e)
+        dispatch(uiBrowseUpdateList({ updateAll: toAdd }))
+        const chunks = chunk(map(toAdd, component => EntitySpec.fromCoordinates(component).toPath(), 100))
+        Promise.all(chunks.map(throat(10, chunk => dispatch(getDefinitionsAction(token, chunk)))))
+          .then(() => uiInfo(dispatch, 'All components have been loaded'))
```
jeffmcaffer: likely only need the error case to notify the user. In the normal case the background loading should be hidden from the user.

src/components/FullDetailView/FullDetailPage.js
```diff
@@ -73,6 +73,7 @@ export class FullDetailPage extends AbstractFullDetailsView {
 
   componentDidMount() {
     const { uiNavigation, component } = this.props
+    console.log(this.props)
```
jeffmcaffer: remove

src/components/Navigation/Pages/PageContribution/ButtonsBar.js
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { Button } from 'react-bootstrap'
+
+export default class ButtonsBar extends Component {
```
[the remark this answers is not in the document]
jeffmcaffer: as discussed, we are deferring the factoring discussion until after this change is in.

src/components/Navigation/Ui/SavePopUp.js
```diff
@@ -0,0 +1,50 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import { Modal, FormGroup, InputGroup, FormControl, Button } from 'react-bootstrap'
+
+export default class SavePopUp extends Component {
+  static propTypes = {
+    prop: PropTypes
```
jeffmcaffer: feels incomplete

src/components/Navigation/Ui/SavePopUp.js
```diff
@@ -0,0 +1,50 @@
...
+              />
+              <InputGroup.Addon>.json</InputGroup.Addon>
+            </InputGroup>
+          </FormGroup>
+        </Modal.Body>
+        <Modal.Footer>
+          <div>
+            <FormGroup className="pull-right">
+              <Button onClick={onHide}>Cancel</Button>
+              <Button bsStyle="success" disabled={!this.state.fileName} type="button" onClick={() => this.doSave()}>
```
jeffmcaffer: could just be `...onClick(this.doSave)`

src/components/Navigation/Ui/SavePopUp.js
```diff
@@ -0,0 +1,50 @@
...
+                type="text"
+                placeholder="Type a name to apply to the file that is going to be saved"
+                onChange={e => this.setState({ fileName: e.target.value })}
+              />
+              <InputGroup.Addon>.json</InputGroup.Addon>
+            </InputGroup>
+          </FormGroup>
+        </Modal.Body>
+        <Modal.Footer>
+          <div>
```
jeffmcaffer: as a point of interest, is the `<div>` needed here?

src/components/UserManagedList.js
```diff
@@ -0,0 +1,271 @@
...
+      })
+    }
+    this.setState({ showVersionSelectorPopup: false }, () =>
+      versions.map(version =>
+        this.onAddComponent(EntitySpec.fromCoordinates({ ...selectedComponent, revision: version }))
+      )
+    )
+  }
+
+  revertAll() {
```
jeffmcaffer: duplicate of superclass method

src/components/UserManagedList.js
```diff
@@ -0,0 +1,271 @@
...
+    this.revert(null, 'Are you sure to revert all the unsaved changes from all the active definitions?')
+  }
+
+  revertDefinition(definition, value) {
+    this.revert(definition, 'Are you sure to revert all the unsaved changes from the selected definition?', value)
+  }
+
```
jeffmcaffer: duplicate of superclass method. There may be more...

src/components/UserManagedList.js
```diff
@@ -0,0 +1,271 @@
...
+  revert(definition, description, value) {
```
jeffmcaffer: suspect that this is (or is largely) duplicate of superclass method. in fact, all the revert logic should be in the superclass as that applies to both user and system managed lists

src/utils/utils.js
```diff
@@ -29,4 +29,32 @@ function asObject(item) {
...
-export { setIfValue, difference, customLicenseIds, asObject }
+const sorts = [
+  { value: 'license', label: 'License' },
+  { value: 'name', label: 'Name' },
+  { value: 'namespace', label: 'Namespace' },
+  { value: 'provider', label: 'Provider' },
+  { value: 'releaseDate', label: 'Release Date' },
+  { value: 'score', label: 'Score' },
+  { value: 'type', label: 'Type' },
+  { value: 'popularity', label: 'Popularity' }
```
jeffmcaffer: do we have this capability? Where is the data to drive that coming from?

src/reducers/uiReducer.js
```diff
@@ -52,17 +52,11 @@ const initialStateNavigation = [
     isSelected: false
   },
   /*{
```
jeffmcaffer: hmmm, not sure. there are some "hidden" routes that we use for deep linking to both the full details PAGE and the page that shows all of a particular PR. Not sure which if any of these is "inspect".

Also, will you need to uncomment this to have Browse show up in the UI?

src/components/UserManagedList.js
storrisi: You are right, i can move all of the rever methods into the System Managed List

src/actions/definitionActions.js
storrisi: That's right, i'm gonna correct this one

storrisi:
I've updated the PR based on your suggestions and also changed the name of the api call method in order to be more generic.

PR #380, opened 2018-11-29
Issue #299 - Sort Spdx License Picker values

### LINKED ISSUE #299: Order of licenses in drop-down menu
With so many SPDX license options, it may be helpful to order them with the OSI-approved license options first (or in order of popularity).  For example, the OSI-approved BSD-3 clause template is listed *last* in the drop-down menu for the BSD-3 clause options.  Same with MIT, I think.  You could also add a link next to each one that would take you to the full SPDX license text for easier reference.

storrisi:
Fixes #299


storrisi:
Thanks, i was using 5.8.0, updating right now.
I've reverted back the package-lock.json file, since i've ran an npm install after the latest updates of the spdx-license-ids library.

PR #383, opened 2018-11-30
Issue #364 - dangling tooltips

### LINKED ISSUE #364: dangling tooltips
From time to time it seems that tooltips are not going away.  In the screenshot below not there are two tooltips. The "Revert" tip is actually stuck up. Even if I move to a different area where there are no tips at all, it stays up, The Copy tip OTOH comes and goes as expected. I've notices this as well quite a bit with the tip for the "Revert Changes" button 

image

Note that I seem to be able to make it happen (in Chrome) by moving the mouse "promptly" over the button. slow enough that the tip pops up but fast enough that you are not lingering.

storrisi:
Fixes #364 

Seems like this is a knows issue of ant-design when there is a Tooltip on a disabled Button, as explained here: ant-design/ant-design#9581

I've tried to replace the Tooltip with the Bootstrap one, but it doesn't work on elements that are not buttons, like a DropDown for example.

As a solution, i've hidden the tooltip if a disabled prop is passed to the render component.


src/components/Navigation/Ui/ButtonWithTooltip.js
```diff
@@ -3,9 +3,9 @@
 import React from 'react'
 import { Tooltip } from 'antd'
 
-const ButtonWithTooltip = ({ button, tip }) => {
+const ButtonWithTooltip = ({ button, tip, disabled }) => {
   return (
-    <Tooltip placement="topLeft" title={tip} arrowPointAtCenter>
+    <Tooltip placement="topLeft" title={!disabled && tip} arrowPointAtCenter>
```
dabutvin: I think this is a nice workaround for the issue!

Can we read the disabled property of `button` or do we need a separate prop for this?

jeffmcaffer: don't think that we can use the disabled prop of the supplied button. In the general case the value fo the `button` prop may not be an instance of the `Button` component so may not have a `disabled` prop. For example, it might be a Button wrapped in some other component (e.g., a Dropzone). So `button` would be an instance of Dropzone.

storrisi:
Actually, i'm not so happy about this workaround as well.
I will investigate on the Bootstrap behaviour in order to find a different and more consistent solution.

storrisi:
I've found what was the issue with Bootstrap, so i've brought it back to the implementation of Tooltips.
Additionally, i've changed the `button` property to more understandable `children`.
Of course this means that instead of passing a button prop to the component, it will render any kind of child of himself.

jeffmcaffer:
@storrisi I am getting an error in ButtonWithTooltip when loading the defintiions page. It seems that `children` is coming in undefined. Was this working for you? Note that I did merge in the latest of master here. I'll continue to look but wanted to check. Perhaps my merge was bad

> React.Children.only expected to receive a single React element child.

storrisi:
It's probably because of the merge of the #375 , i'm gonna update this one since something has changed

storrisi:
@jeffmcaffer fixed, now it should be fine, thanks for catching up

PR #387, opened 2018-12-03
Issue #368 - Fixing behaviour of Dropdown button

### LINKED ISSUE #368: change/add versions dropdown behavior is odd
To fix some expansion problems `stopPropagation()` was added to the handler the dropdown expand handler for the definition entry "change/add versions" dropdown. This addressed the most obvious problem but feels wrong and does not actually work correctly. With that change clicking on the button does not expand/collapse the menu like it should. Without the change, the menu behaves correctly but the row also expands collapses.

The DropDown component really should be consuming this itself. Perhaps there is a setting on the Antd dropdown?

storrisi:
This PR should fix the behaviour mentioned in #368 and initially solved in #365 
Now the dropdown shows up correctly when clicking.

Fixes #368


storrisi:
Because of the implementation in #383 , this one needs to be updated when the mentioned PR will be merged

PR #393, opened 2018-12-04
Issue #389 - Added specific store management method

### LINKED ISSUE #389: Curation entries are not expandable
This was a regression in the latest master

https://dev.clearlydefined.io/curations/230

The issue is clicking on the entry does not expand it anymore

The workaround is to click the magnifying glass to view the full detail

@capfei fyi

image

storrisi:
Fixes #389


dabutvin:
that's unfortunate. I can look into it and see if it's something we can fix easily - or swap netlify out for a different one of these providers

PR #400, opened 2018-12-07
LicensePicker component

### LINKED ISSUE #255: UI should allow for adding multiple declared licenses
Right now the Browse UI for editing components only provides one drop down and only allows you to add one declared license. For components that have multiple licenses, we should enable curators to add multiple licenses to the component.

storrisi:
This is an initial and WIP activity that fixes #255  , based on https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60

Some improvements are still needed to the UI, as well as implement grouped expressions.


storrisi:
@jeffmcaffer here's a quick video of the behavior.
As i mentioned, the graphic is still ugly, and some functions are still missing, but you can start see what is the flow that we have in mind.

dec-14-2018 17-30-57

storrisi:
@jeffmcaffer did you had the chance to take a look to the behavior explained in the video? 
I have some other changes ready to be applied to this one, but a feedback from you would be helpful.

Thanks

jeffmcaffer:
Thanks for the video. seems like a good start and we can put this in now as it is definitely an improvement over what we have today. Some questions:

* How are () handled?
* How does the user get to this? Is this the only experience for editing license entries?
* Can we make this much smaller? more like a pop up/over than a full on, full screen dialog? Just would feel lighter weight for the user.
* How will this mesh with the Suggestions experience
* Hard to tell in the video, are you clicking on Add new group to get a new entry? If so, perhaps just a + icons?
* Is it possible to delete a group?

BTW, did you look at the existing "query builders" like https://querybuilder.js.org/ or https://www.npmjs.com/package/react-querybuilder? Are they useful in this context?

storrisi:
Thanks for your feedback @jeffmcaffer , i'm going to put in place some changes to this code that i've already made.

>How are () handled?
In this version wasn't possible to handle groups, i've improved the code meanwhile and i will update soon the PR with that functionality, and also sharing a new video which shows the behavior

>How does the user get to this? Is this the only experience for editing license entries?
We can give to the user a choice, type in manually or use the logical builder. But in both ways i would suggest to keep the validation tool that shows to the user if the typed license is a correct one.

>Can we make this much smaller? more like a pop up/over than a full on, full screen dialog? Just would feel lighter weight for the user.
Sure

>How will this mesh with the Suggestions experience
The suggestion process is something that doesn't affect how a user build a new license.
I will provide a video also for that, but basically if there is a suggestion the user can select it from a list coming from the server. Then, maybe he wants to make some little changes and he can edit the license himself.

>Hard to tell in the video, are you clicking on Add new group to get a new entry? If so, perhaps just a + icons?
For "plain" licenses, there is an automation: if the user select a license, and then choose a conjunction from the list, a new element appears.
To add a new group the user should do a manual activity clicking on the button.
That would build a new set of elements under the parent one.

> Is it possible to delete a group?
Not yet, i'm working on editing and deleting functionalities.

>BTW, did you look at the existing "query builders" like https://querybuilder.js.org/ or https://www.npmjs.com/package/react-querybuilder? Are they useful in this context?
Those are exactly my starting point on building this

pombredanne:
@storrisi Thank you. I am looking at https://github.com/clearlydefined/website/pull/400#issuecomment-447378857 and I feel like a single text field with editing/completion assistance might be easier to use for me: fewer clicks, fewer context switch, easier to edit/delete and I can see the whole expression I am building rather than bits of it.

storrisi:
@jeffmcaffer i've finalized the work here. Now the Picker handle any kind of rules for generating or modify a License, based on the SPDX License rules.
I will provide a video of the updated behavior shortly.

Please note that the folder structure of this component has been structured to be standalone-ready.
While working on Standalone Components, i will provide a change to this one as well.

src/components/LicensePicker/RuleBuilder.js
```diff
@@ -0,0 +1,122 @@
...
+    removeRule: PropTypes.func //callback function called when removing a single license or a group
+  }
+  renderRule = (rule, path, conjunction = null, parentRule = {}) => {
+    const { changeRulesOperator, updateLicense, considerLaterVersions, addNewGroup, removeRule } = this.props
+    const currentPath = path[path.length - 1]
+    if (rule.license || rule.license === '')
+      return (
+        <Col md={12} style={{ padding: '10px' }}>
+          {currentPath !== 'right' && (!parentRule.hasOwnProperty('left') && !parentRule.hasOwnProperty('right')) ? (
+            <Col md={12} style={{ padding: '0px', display: 'flex', justifyContent: 'space-between' }}>
```
jeffmcaffer: Can this style be factored out into a css class? Seems like a duplicate of L67

src/components/LicensePicker/RuleBuilder.js
```diff
@@ -0,0 +1,122 @@
...
+                  OR
+                </ToggleButton>
+              </ToggleButtonGroup>
+              <div>
+                <Button onClick={() => changeRulesOperator('AND', path)}>Add Rule</Button>
+                <Button onClick={() => addNewGroup(path)}>Add Group</Button>
+              </div>
+            </Col>
+          ) : null}
+          <Col md={4} style={{ display: 'flex', alignItems: 'center' }}>
```
jeffmcaffer: factor out style to a reusable class. This just centers things. We will do that all over in general. Even if we can't share the .css file between independent components, it is easier to duplicate the class and use that every where than paste in the style everywhere.

jeffmcaffer: Suggest factoring this out into a separate function to support readability. The reader has to dissect all of this to understand what UI element it represents.

src/components/LicensePicker/RuleBuilder.js
```diff
@@ -0,0 +1,122 @@
...
+            <SpdxPicker value={rule.license} onChange={value => updateLicense(value, path)} />
+            {rule.license !== '' && (
```
jeffmcaffer: is `rule.license` better here? The current code will be `true` if `rule.license === null`. is that what you want?

src/components/LicensePicker/RuleBuilder.js
```diff
@@ -0,0 +1,122 @@
...
+                  Any later version
+                </div>
+              </Fragment>
+            )}
+            {path.length > 0 && <Button onClick={() => removeRule(path)}>x</Button>}
+          </Col>
+        </Col>
+      )
+    return (
+      <Col md={12} style={{ padding: '10px', paddingRight: '0px', background: 'lightgrey', border: '1px solid black' }}>
```
jeffmcaffer: more style comments. Factor out for readability and reuse.

jeffmcaffer: also consider factoring this in to a function as well.

src/components/LicensePicker/__tests__/LicensePicker.test.js
```diff
@@ -0,0 +1,225 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import LicensePicker from '../LicensePicker'
+import LicensePickerUtils from '../utils'
+
+const license = 'Apache-2.0 AND MIT'
+describe('LicensePicker', () => {
```
jeffmcaffer: LOVE this! TESTS! yay!

src/components/LicensePicker/style.css
```diff
@@ -0,0 +1,25 @@
+.spdx-picker {
+  font-family: 'Montserrat', sans-serif !important;
```
jeffmcaffer: What the style pluggability story for independent components? If this were used in a different site that did not use Montserrat, how would they change. Related, we are likely to change the font of the clearlydefined site. can this be inherited/supplied somehow by the outer context?

src/components/LicensePicker/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.spdx-picker-header-title {
```
jeffmcaffer: I'm always torn on this style naming and reuse. On the one hand, this name is awesomely specific and makes it very clear that it drives the style of the picker header. OTOH, lots of things need to be centered and there are many ways of centering. Having that technique defined in one place and shared is awesome. Thoughts?

src/components/LicensePicker/style.css
```diff
@@ -0,0 +1,25 @@
+.spdx-picker {
```
jeffmcaffer: meta point here, this does css directly rather than scss? That means we'll not get the shared changes and the styling approaches are different. Thoughts on a direction here?

src/components/LicensePicker/utils.js
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+import parse from 'spdx-expression-parse'
```
jeffmcaffer: @dabutvin did a bunch of work on parsing and validating SPDX expressions to better suit the needs of ClearlyDefined (e.g., accommodating NOASSERTION). We should look at if/how that is applicable here.

src/components/LicensePicker/utils.js
```diff
@@ -0,0 +1,124 @@
...
+const NOASSERTION = 'NOASSERTION'
+
+// Shared methods appliable to License Picker
+export default class LicensePickerUtils {
+  static parseLicense(license) {
+    return license ? parse(license) : {}
+  }
+
+  // Returns a license string based on the rules in input, following the specification of https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60
+  static stringify(obj) {
```
jeffmcaffer: suggest `toString`.  also better name for `obj`. Here it seems to be `expression`?

Here and elsewhere

src/components/LicensePicker/utils.js
```diff
@@ -0,0 +1,124 @@
...
+    if (!obj) return null
+    if (obj.hasOwnProperty('noassertion')) return NOASSERTION
+    if (obj.license) return `${obj.license}${obj.plus ? '+' : ''}${obj.exception ? ` WITH ${obj.exception}` : ''}`
+    const left =
+      obj.left && obj.left.conjunction && obj.left.conjunction.toLowerCase() === 'or'
```
jeffmcaffer: `get(obj, 'left.conjunction', '').toLowerCase() === 'or'`

here and lower down

src/components/LicensePicker/utils.js
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
```
jeffmcaffer: This file is really a defining an spdx expression class. In other places we have a mess of utilities etc for dealing with expressions. it would be useful to gather all of these together into one class and put that out as an npm.

Don't need to do anything for this PR but it would be an interesting thing to do. @dabutvin, what do you think?

src/components/LicensePicker/utils.js
```diff
@@ -0,0 +1,124 @@
...
+   * @param  {} obj current object rules
+   * @param  {} path where apply the change
+   * @returns updated object rules
+   */
+  static createRules(conjunction, obj, path) {
+    if (path.length > 1 && obj.hasOwnProperty(path[0]) && obj[path[0]].hasOwnProperty(path[1])) {
+      obj[path[0]] = this.createRules(conjunction, obj[path[0]], path.slice(1))
+      return obj
+    }
+    if (path.length === 0) return this.createRuleObject(conjunction, obj.left || obj, obj.right && obj.right)
```
jeffmcaffer: `obj.right && orbj.right` will just give a boolean. (end of line). should this be `obj.right || obj`? or just `obj.right`? or `!!obj.right`?

src/components/LicensePicker/utils.js
```diff
@@ -0,0 +1,124 @@
...
+          ? this.createRuleObject(conjunction, obj.right.left, obj.right.right)
+          : obj.conjunction === conjunction && this.createRuleObject(conjunction, obj.right)
+
+    return this.createRuleObject(ruleConjunction, left, right)
+  }
+
+  static createGroupObject() {
+    return {
+      left: { license: '' },
+      conjunction: 'and',
```
jeffmcaffer: why is it always 'and'?

src/components/LicensePicker/utils.js
```diff
@@ -0,0 +1,124 @@
...
+// SPDX-License-Identifier: MIT
+import parse from 'spdx-expression-parse'
+import isNil from 'lodash/isNil'
+import unset from 'lodash/unset'
+import get from 'lodash/get'
+import set from 'lodash/set'
+const NOASSERTION = 'NOASSERTION'
+
+// Shared methods appliable to License Picker
+export default class LicensePickerUtils {
```
jeffmcaffer: Nice set of utils. The logic here seems complicated enough that it needs a boatload of tests. Some pretty subtle stuff happening.

jeffmcaffer:
Would love to see the video @storrisi

src/components/LicensePicker/style.css
storrisi: It depends on the specific behavior. In this case, since the style applies only a center, we can define some general classes and use that ones.
But given this could be a standalone component, we should leave also a custom class in order to let the user to customize it as he wants.

src/components/LicensePicker/style.css
storrisi: Given this could be a shared component, i preferred to use pure css instead of scss.
Maybe in some project users couldn't have implemented that preprocessor.
For our specific case, we can create a specific scss file that handles the specific custom style for this component.
What do you think?

src/components/LicensePicker/utils.js
storrisi: Thanks for pointing this out, i will take a look at the service to see how that could be used.

src/components/LicensePicker/utils.js
storrisi: @jeffmcaffer i agree, specially if some of these things are shared with the service too (like the parse method)

src/components/LicensePicker/utils.js
storrisi: While adding a new group object, the user doesn't know yet what could be the joining conjunction of the left and right element. So, in order to create an empty object that could be parsed from the parse method, i've set a default 'and' conjunction.

src/components/LicensePicker/RuleBuilder.js
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+import React, { Component, Fragment } from 'react'
+import PropTypes from 'prop-types'
+import { ToggleButtonGroup, ToggleButton, Button, Row, Col } from 'react-bootstrap'
+import SpdxPicker from '../SpdxPicker'
+
+export default class RuleBuilder extends Component {
+  constructor(props) {
```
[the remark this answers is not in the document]
storrisi: You are right, don't know why it didn't worked.

storrisi:
@jeffmcaffer thanks for the feedback.
On the first attempt i didn't applied so much effort on styling, that's way i used some style props instead of building a css class.
I've also fixed some methods based on your suggestions.

I'm gonna upload a video in a while.

storrisi:
senzanome

PR #402, opened 2018-12-14
Fix for change of schema definition

storrisi:
Fixes a bug introduced by https://github.com/clearlydefined/service/pull/323 related to the `_meta` property of the definition schema


jeffmcaffer:
These other properties should be added to the schema for `definitions`. We are trying to be pretty strict about keeping valid docs at all times

dabutvin:
👍 👍

PR #409, opened 2019-01-11
Standalone spdx picker

storrisi:
This follows #400 , starting from the PROPOSAL for standalone components #313 i've structured the LicensePicker as a standalone component.
I've also updated the README files with the development process.

We should check if everything works during the CI process.

If something will fail, i suggest to give [Lerna](https://github.com/lerna/lerna) a try.


storrisi:
As i expected, the CI failed.
Looks like `azure-pipelines.yml` needs to be updated in order to run an `npm install `in the `components` folder.
I'm not so experienced on how azure pipelines configuration works, i will take a look at this.
@jeffmcaffer do you have any suggestion?

storrisi:
This was only a test to switch the LicensePicker as a standalone component following the PROPOSAL #313 

It can be closed for now.

PR #413, opened 2019-01-21
Add notice file sharing to Definitions

jeffmcaffer:
The service can now generate Notice files so add a simple way of letting users create notices from the Definitions list.


src/components/UserManagedList.js
```diff
@@ -167,36 +168,46 @@ export default class UserManagedList extends SystemManagedList {
...
-  async createGist(name, content) {
+  async saveNotices(coordinates, options) {
+    const { token, dispatch } = this.props
+    uiInfo(dispatch, `Creating notice file for ${coordinates.length} coordinates...`)
+    const list = coordinates.map(entry => entry.toString())
+    const notices = await getNotices(token, list, options.renderer, options.options)
+    const { summary, content } = notices
+    const message = `Created Notices file with ${summary.total} entries, ${summary.missing} missing entries, and ${
+      summary.warnings.noLicense.length
+    } missing licenses`
```
dabutvin: let's add a link in this toast to fix the ones with missing licenses.
This will filter the list to 'absence of license' so they can be curated

Issue #363, opened 2018-11-11
testing setup improvements


jeffmcaffer:
In the current setup we have `npm test` which runs eslint and `react-scripts test`. Some issues
* eslint does not seem to find all the issues. We just had about 10 different warning when using `npm start` but none of them showed up in eslint output
* The test entry point runs all of the tests including the expensive component tests that take many seconds. Bummer when you are trying to write unit tests for utilities etc.
* The test framework seems to use a different version of `expect` etc. in all the other repos for ClearlyDefined we use expressions like `expect().to.be.equal` etc. Seems that in the website setup `be` is not a known thing. Ideally we'd use the same version/setup as all the other repos. If that is not possible for some reason, it would be great to know what `expect` api we are using 
* need a vscode launch config for running the tests so that we can easily debug with breakpoints

jeffmcaffer:
@storrisi lets figure this one out and drive to close

storrisi:
> eslint does not seem to find all the issues. We just had about 10 different warning when using npm start but none of them showed up in eslint output

Looks like that we are not able to update `jest` configuration since we are running a `create-react-app` application. But i will investigate further more on this topic.

> The test entry point runs all of the tests including the expensive component tests that take many seconds. Bummer when you are trying to write unit tests for utilities etc

I think that is due of running tests with the `--coverage` option.
Without it, all the tests will run only for the first time, after that the process will not close and waits for any change, running only tests for modified components or tests as well.

> The test framework seems to use a different version of expect etc. in all the other repos for ClearlyDefined we use expressions like expect().to.be.equal etc. Seems that in the website setup be is not a known thing. Ideally we'd use the same version/setup as all the other repos. If that is not possible for some reason, it would be great to know what expect api we are using

We are using https://jestjs.io/docs/en/expect , which is a different assertion compared to https://www.chaijs.com/api/bdd/ which is used in the service

> need a vscode launch config for running the tests so that we can easily debug with breakpoints

Do you mean something like https://github.com/jest-community/vscode-jest or https://marketplace.visualstudio.com/items?itemName=firsttris.vscode-jest-runner?

storrisi:
> The problem is that without --coverage the process does not exit and we cannot us this in CI builds. My point with the original comment was that we could/should separate the tests so that we can run the fast unit tests as we are developing code and then the longer integration tests on demand. The CI should run them all.

Got it.  I'm gonna investigate on this.

[...]

PR #415, opened 2019-01-22
[PROPOSAL] Unit and Integration test separation

storrisi:
This is a first attempt to make a separation to run unit and integration tests separately as mentioned in #363 
It's based directly on jest , running different configuration files based on the specific behavior.
The separation is done giving a different naming to files that contains only unit tests, which could be tested running a specific `test:unit` or `test:integration` script
The main `test` script hasn't been changed.

Unfortunately there is a issue while running integration tests for files that import css files from third-party libraries (like react-bootstrap)


storrisi:
I've updated the PR removed css files from the `jest` setup.
This allows the test to run without any problem.

PR #416, opened 2019-01-23
E2E tests improvement

### LINKED ISSUE #397: Define e2e tests specs
We need to figure out what kind of e2e tests we want to write for the website.

Those should be something like user stories ("As a user i want to go to a page and click on some elements and return a result") and should cover the most critical parts of the application.

storrisi:
This is an improvement of the previous implementation of E2E tests.
As discusse, now are used more readable class names instead of DOM selectors.

Also, tests have been updated to support the use of the new License Picker.

Fixes #397


storrisi:
@jeffmcaffer  as mentioned in #397 i've defined the steps to test a contribution flow.
The test creates a contribution using the dev APIs and an anonymous user.
I've found a problem with the Source Picker while working on this, i'm going to investigate on what is wrong and then post a fix.

src/components/DefinitionEntry.js
```diff
@@ -240,8 +240,8 @@ export default class DefinitionEntry extends React.Component {
             <Col md={10} className="definition__line">
               {this.renderWithToolTipIfDifferent(
                 'licensed.declared',
-                <InlineEditor
-                  field={'licensed.declared'}
+                <ModalEditor
+                  field={'licensed-declared'}
```
jeffmcaffer: Hmmm, from a functional point of view it would be better to use '.' as we are talking about addressing javascript objects (like lodash does). The fact that this is sometimes used as a style name is an implementation detail. Alternatively, the prop should not be called `field` and perhaps be something `style` related (though like bootstrap, perhaps not exactly `style`)

src/components/InfiniteList.js
```diff
@@ -59,7 +59,7 @@ export default class InfiniteList extends React.Component {
             {({ width }) => (
               <List
                 ref={this.hookRef(registerChild)}
-                className={styles.List}
+                className={`${styles.List && styles.List} ${customClassName}`}
```
jeffmcaffer: What do you mean here? `` `${styles.List && styles.List}` `` === `` `${styles.List}` `` I believe

src/components/ContributePrompt.js
```diff
@@ -79,7 +79,7 @@ export default class ContributePrompt extends Component {
                     {session.isAnonymous ? 'anonymous' : `@${session.username}`}
                   </FormControl.Static>{' '}
                   {session.isAnonymous && (
-                    <Button bsStyle="success" onClick={onLogin}>
+                    <Button bsStyle="success" className="contributePrompt__modal--loginButton" onClick={onLogin}>
```
jeffmcaffer: Should likely note the BEM naming model use in the coding style doc somewhere.

Should we also tweak these names to indicate that they are not "real" styles? AFAICT we would never want to actually have a style called that. Rather we are adding these for identification/testing purposes. My expectation is that our styling is more macro with rare exceptions where we tweak just one element. For example, we'd figure out the styling for a general class of UI scenarios and then apply that everywhere (e.g., "spaced-out-list") rather than driving the style of each instance individually.

Make sense?

src/components/Navigation/Sections/HeaderSection.js
```diff
@@ -50,18 +50,23 @@ export default class HeaderSection extends Component {
         <Col md={4} className="text-right">
           {!isEmpty(changes) && (
             <ButtonWithTooltip tip="Revert all changes of the current definition">
-              <Button bsStyle="danger" onClick={() => handleRevert()}>
+              <Button bsStyle="danger" className="revertButton" onClick={() => handleRevert()}>
```
jeffmcaffer: CSS classes seem to shy away from camelCase and go towards `-` (i.e., revert-button) and we do that elsewhere. 

Like above, I think we should distinguish between "real" and "identification" classes.

src/components/DefinitionEntry.js
storrisi: We can use the dot separator, even if the `field` prop is used like a className, like in this case.
There's only a workaround to apply if someone needs to apply a css class to that element, and it's to prefix the dot with a `\` in the css file.

Ex:
```
.licensed\.declared{

}
```

src/components/InfiniteList.js
storrisi: It should render the `styles.List `value only if exists and it's not undefined.
When using the `` notation, any undefined value would be printed anyway.

src/components/DefinitionEntry.js
storrisi: Looks like the tests could have some problems while trying to identify a className with this kind of notation.
I've converted all of them to a `name` attribute.

src/components/ContributePrompt.js
storrisi: I got your point, it depends if we want to use this identifiers only for testing purposes or to attach a real class that can be customized.
The BEM notation helps to identify some specific elements (ex. we can have multiple loginButton in the same view, but i want to test only the one in the contribute modal).

We can address the same thing using different approaches, like using ids or attributes (like name).
We should keep a strategy and go on.

Thoughts?

jpeddicord: Re: discord conversation:

This class name itself is a little wild, but it's obvious enough visually that I'd not personally try to style something against. Class names are a good identifier for that purpose, and don't _necessarily_ have semantic value. It does put some extra data into the DOM, but it's very minimal.

If you're looking for an XPath approach, selectors like this could be useful:

```
//div[@class="modal"]//button[text()="Login"]
```

This would select a button with the text of "Login" within a div (anywhere in the DOM) with a class of "modal". `//` is essentially a relative search; you can think of it as `/**/` with glob-style patterns.

Note that `class` doesn't have any special meaning in an XPath query; it's one string and not a set of classes. A more "correct" expression would be:

```
//div[contains(concat(' ',normalize-space(@class),' '), 'modal']//button[text()="Login"]
```

That allows "modal" as one of the classes, not the only class. I didn't verify this exact expression against the code in this PR; but hopefully you get the idea.

Great XPath resource here: https://devhints.io/xpath and your browser console can likely evaluate these as well. In Firefox `$x` is a function to evaluate/play around with expressions in the console. So yeah, it's kinda gnarly. XPath expressions are super useful..... but you could also get away just fine with CSS classnames like this one. :)

jeffmcaffer: So from the above and some of the discussion in Discord, using the classes seems easy and attainable. The BEM style names are pretty verbose but its a known convention so why invent another unless we need to. The only tweak I'd put on is *perhaps* a `t_` prefix or something to indicate that this is not actually a class, just a name. 

While I agree with @jpeddicord that the names here are unattractive enough that folks would likely not try to style, I worry about test oriented names higher up in the dom. For example, `contributePrompt`. (well, `contribute-prompt`) That is attractive as an actual style class. `t_contribute-prompt` (or something) is different enough to give pause.

Having said all of that, it feels "yucky" to be sprinkling these long class names all over the code. https://ghostinspector.com/blog/css-selector-strategies-automated-browser-testing/ has an interesting hierarchy in the "How do I know when to use the different methods?"How do I know when to use the different methods? section that perhaps we can use to guide us here. The main point here is to use something reasonably stable. Something that is already there would be great.

storrisi: I've read the article and it sounds fair enough. I will try to change some identifiers logic in order to have less specific class names that could create some confusion to the developers.

storrisi:
I've updated tests based on the discussion and articles about the identifiers, using most of the DOM elements and identifiers that are already existing in the code.
Somewhere i've used some specific ids or classes to identify better the element.

I've also switched the controls using `jest-puppeteer` assertions, because `puppeteer`, if used alone, doesn't return any significant error while failing (basically returns only a timeout without specifying where and why).
`jest-puppeteer` assertions let us to have a more specific detail in case of errors.

src/components/ContributePrompt.js
storrisi: I've applied most of the rules, but, anyway, a point to consider is that if a user wants to edit the style of a specific element, that could be done even without any specific classname, using attributes or ids, or style prop directly.
That is, using any existing element or class for tests purposes avoids to add other code to maintain, and that's perfectly fine, but doesn't discourage a user to customize the style, if he really wants.

jeffmcaffer: Oh absolutely. My point was not about "stopping" people doing it, rather not encouraging them to. If you walk up to a bit o DOM and it has some css classes identified, it's pretty reasonable to think "hey, I could scoot that element over a bit with a margin, or change the color by tweaking that class".  

Additionally, we are putting a class name there but don't actually have a class. It's good to point out that that is intentional. Otherwise I could see someone cleaning up and removing "unsued" classes.

It's just about making the code more self-evident.

src/components/ContributePrompt.js
```diff
@@ -79,7 +79,7 @@ export default class ContributePrompt extends Component {
                     {session.isAnonymous ? 'anonymous' : `@${session.username}`}
                   </FormControl.Static>{' '}
                   {session.isAnonymous && (
-                    <Button bsStyle="success" onClick={onLogin}>
+                    <Button bsStyle="success" className="login-button" onClick={onLogin}>
```
jeffmcaffer: Here I wonder why we use `className` as opposed to `id` as on L68. I'm not super fussed one way or the other (have a slight preference for `id` as that's "a thing") but would like to understand the thought process so we can have consistency in the codebase. Similar comment for elements below.

One way to look at is that major elements having `ids` is pretty natural and that can be used for a number of different scenarios. But following some broader convention would be interesting.

@jpeddicord, @dabutvin  thoughts?

src/components/Navigation/Sections/HeaderSection.js
```diff
@@ -50,18 +50,23 @@ export default class HeaderSection extends Component {
...
-              <Button bsStyle="danger" onClick={() => handleRevert()}>
+              <Button bsStyle="danger" className="revert-button" onClick={() => handleRevert()}>
                 <i className="fas fa-undo" />
                 <span>&nbsp;Revert Changes</span>
               </Button>
             </ButtonWithTooltip>
           )}{' '}
           {modalView && (
-            <Button bsStyle="primary" disabled={isEmpty(changes)} onClick={handleSave}>
+            <Button bsStyle="primary" className="save-button" disabled={isEmpty(changes)} onClick={handleSave}>
```
jeffmcaffer: is this `save-button` or `ok-button`? That would match `cancel-button`

src/components/Navigation/Ui/ComponentButtons.js
```diff
@@ -108,10 +108,16 @@ export default class ComponentButtons extends Component {
                   trigger={['click']}
                   overlay={
                     <Menu>
-                      <Menu.Item onClick={this.showVersionSelectorPopup.bind(this, currentComponent, false)}>
+                      <Menu.Item
+                        className="switchComponentVersion"
```
jeffmcaffer: camelCase? should this be `switch-component-version`?  similarly below

src/components/SpdxPicker.js
```diff
@@ -39,6 +39,7 @@ export default class SpdxPicker extends Component {
     return (
       <div className="editable-editor">
         <Typeahead
+          className="spdx-input-picker"
```
jeffmcaffer: `id`?

src/components/SourcePicker.js
```diff
@@ -67,7 +67,7 @@ class SourcePicker extends Component {
     const { activeProvider, selectedComponent } = this.state
     const { value } = this.props
     return (
-      <Grid className="main-container">
+      <Grid className="main-container" id="source-picker">
```
jeffmcaffer: this feels natural (id is the id, classnames are actual styles)

src/components/ContributePrompt.js
storrisi: Id should be used only when a DOM element is unique over the page. In this case, a generic Login Button could be present somewhere else in the application, even if actually this doesn't happen.
That's why i kept the className attribute instead of switching it to a unique id.

At Line 68 i've used an `id` since the contribute modal is really unique, there can't be another contribute modal at the same time.

src/components/SpdxPicker.js
storrisi: Applying an `id` is not possible in this case.
The `SpdxPicker` is used multiple times in the application (like in the `LicensePicker` for example) so an id would generate some conflicts in the DOM.

src/components/SourcePicker.js
storrisi: Since the component is really unique over the application, that's the correct approach.

storrisi:
@jeffmcaffer i've updated the PR adding a Page mapping object used in the test. This should be useful to make some other discussions.
I've found this approach very powerful while working, it helps to keep the test agnostic over the page structure.

jeffmcaffer:
thanks @storrisi. the mapping object makes sense as a good way of deduplicating the code. Thanks for the example. Can we augment that with something like the data custom props to make the map entries themselves resilient. 

So, something like this 
* use the mapping objects
* in the mapping objects use the move prevalent existing id, name, class, ... 
* in the event that no such id, name, ... is available, use `data-test-id` custom props to create such a selector, to make the most resilient "path"

storrisi:
I've added some data attributes where necessary, removing some classNames that i've added before.
I think that now this approach could be considered a good guideline for future E2E tests.

jeffmcaffer:
Thanks @storrisi. in the current code i still see a number of new `className` additions. should these all be data-test-id (or some such). It seems much clearer if classNames are CSS / style things and test stuff is in data-test-id attributes.

storrisi:
I've switched all the applied `classNames` to `data-test-id` attributes.
Unfortunately today the `contribution` API in the dev environment is failing with a 400 generic error, which cause the tests to fail.

On a side note:
in components like `Typeahead` is not possible to use a custom `data-test-id` attribute, since it's not applied to any of the rendered DOM elements.

storrisi:
> Unfortunately today the `contribution` API in the dev environment is failing with a 400 generic error, which cause the tests to fail.

This has been fixed by @dabutvin so now all the tests are successful.

storrisi:
@jeffmcaffer do you think this one is ready to go or it's required to implement something else?

dabutvin:
@storrisi see https://github.com/clearlydefined/website/pull/467#issuecomment-463727439

After we get this up to date, let's get it merged!

storrisi:
I've updated the tests in relation of the new behavior of the License field item.

dabutvin:
I'm getting failures running `npm run e2e:test`

for example: 

```
Home page › Home page sections › should have a "ClearlySecure" section
```

do you think we should get all the tests running under `npm test` ?

storrisi:
@dabutvin did you ran `e2e:start` before? the `e2e:test` script checks for `localhost:3333`, so it needs the application running on that port.

> do you think we should get all the tests running under `npm test` ?

I don't think it could be the best way from a developer point of view, since that script would run all the tests.
If a user is testing a single class or component, he don't want to spend so much time waiting for all the tests, even e2e, run and pass.

Perhaps it could be better to have a specific script for the CI, that would run ALL the tests.
What do you think?

dabutvin:
I see. I'll give it a try and see how it goes.

We can probably leave the setup as is and modify the city to run all the tests

dabutvin:
the tests only pass in debug mode right now

storrisi:
> the tests only pass in debug mode right now

@dabutvin it should be fixed now

dabutvin:
alright! let's land this.
I'll work on the CI updates required so we can stay green here

PR #425, opened 2019-01-30
allow user to filter on Notice file problems

jeffmcaffer:
Show a `Filter` button when there are errors generating a Notice file. this is the first step to having a real Notice file experience.


jeffmcaffer:
@dabutvin I'd like to get this to production along with any other notice related code in time for demos tomorrow CET. can you take a look?

src/components/SystemManagedList.js
```diff
@@ -207,11 +207,15 @@ export default class SystemManagedList extends Component {
...
 
-  onFilter(value) {
-    const activeFilters = Object.assign({}, this.state.activeFilters)
-    const filterValue = get(activeFilters, value.type)
-    if (filterValue && activeFilters[value.type] === value.value) delete activeFilters[value.type]
-    else activeFilters[value.type] = value.value
+  // note that in some scenarios `onFilter` is called with a random second arg
+  // and sometimes with an explicit overwrite intent (e.g., true)
+  onFilter(value, overwrite = false) {
+    const activeFilters = overwrite === true ? value : Object.assign({}, this.state.activeFilters)
```
storrisi: Could `overwrite` have other non-boolean values? If no, this check could become `overwrite ?`

src/components/SystemManagedList.js
```diff
@@ -207,11 +207,15 @@ export default class SystemManagedList extends Component {
...
+    if (overwrite !== true) {
```
storrisi: Same here, could be better `if(!overwrite)` ?

src/components/UserManagedList.js
```diff
@@ -174,8 +175,8 @@ export default class UserManagedList extends SystemManagedList {
   async doSave() {
     const { components } = this.props
     const spec = this.buildSaveSpec(components.list)
-    await this.saveSpec(spec)
     this.setState({ showSavePopup: false })
+    this.saveSpec(spec)
```
storrisi: If this method needs to be called after the state change, then is better to use the callback method of the `setState()` function:
`this.setState({ showSavePopup: false }, () => this.saveSpec(spec))`

dabutvin:
should there be a way to remove the activeFilter?

src/components/UserManagedList.js
```diff
@@ -199,14 +200,25 @@ export default class UserManagedList extends SystemManagedList {
...
     else uiInfo(dispatch, message)
     return saveAs(new File([content], `${options.filename}`))
   }
 
+  _renderNoticeProblem(baseMessage, summary) {
+    return (
+      <div>
+        {`${baseMessage}, and ${summary.warnings.noLicense.length} missing licenses`}
+        <div>
+          <Button bsStyle="warning" onClick={() => this.onFilter({ 'licensed.declared': 'absence' }, true)}>
```
dabutvin: it would be great to get `{ 'licensed.declared': 'absence' }` to  include NONE and NOASSERTION.

opened https://github.com/clearlydefined/website/issues/428

PR #440, opened 2019-01-31
handle "no value" case in list filtering => prod

jeffmcaffer:


src/components/SystemManagedList.js
```diff
@@ -196,7 +196,7 @@ export default class SystemManagedList extends Component {
         if (value === 'presence') {
           if (!fieldValue) return false
         } else if (value === 'absence') {
-          if (!['NONE', 'NOASSERTION'].includes(fieldValue)) return false
```
dabutvin: 👍 was just about to do this

PR #441, opened 2019-01-31
highlight problem score elements

jeffmcaffer:
Make the score elements yellow or red depending on how far off their full score they are.


dabutvin:
very cool 👍

Issue #412, opened 2019-01-20
license picker gets in odd state


jeffmcaffer:
I don't see what is wrong with expression but the bar at the top is red... Theoretically if I just click buttons and fill in all the blanks the expression should always be syntactically valid no?

![image](https://user-images.githubusercontent.com/10070956/51443465-0b35da80-1c9e-11e9-8086-02d802b71308.png)

storrisi:
After each change the expression is passed to the `spdx-expression-validate` library, which returns if the passed expression is a valid license.

Maybe those group of licenses are not possible for some SPDX rules?

jeffmcaffer:
AFAIK the expression validation is not doing any license interpretation. It is analyzing the structure and that each license id is valid. So it should be impossible for someone to fill in all of the boxes and have an invalid expression. 

Perhaps the conversion from the structure here to a string expression is messed up? The validator may be needing `()`s around some of the ANDs (not sure).

One related point, I'm not sure that all licenses allow for `+` (any later version). @dabutvin can you confirm here?

storrisi:
I see, probably you mean that the Picker in this situation should return `(MIT AND ADSL) AND (AFL-3.0 OR AGPL-1.0-only)`

I'm gonna investigate on this.

storrisi:
I've tried to switch on the Picker the last license from `AGPL-1.0-only `to `0BSD` and the validation library returns `true` for `MIT AND ADSL AND (AFL-3.0 OR 0BSD)` and `false` for `MIT AND ADSL AND (AFL-3.0 OR AGPL-1.0-only)`

So i guess that the `spdx-expression-validate` library applies some kind of validation of the used licenses

storrisi:
@jeffmcaffer do you have any update on this?
AFAIK , as i reported in my last comment, looks like that the `spdx-expression-validate` is validating also the used license.
It happens also when you choose a `MIT-0` as a main license.

storrisi:
While applying a `MIT-0` license, the message returned is `"Lexical error on line 1. Unrecognized text.↵MIT-0↵---^"`

dabutvin:
the only thing spdx-expression-validatate does is a try/catch around spdx-expression-parse
https://github.com/jslicense/spdx-expression-validate.js/blob/master/index.js

if it is parseable, then it is validated.

storrisi:
I'm trying to debug the library, but i can't understand why and when that error is returned

storrisi:
[...]

PR #452, opened 2019-02-05
Test for License validation

storrisi:
I've wrote a test to address the problem of validation mentioned in #412 

The test is failing when the library tries to validate a 'MIT-0' license, while passing without any issue if checking 'MIT-CMU'


dabutvin:
I just had a look at this @storrisi and it looks like MIT-0 is a new license IDs added to spdx-license-ids after 3.0.0

https://github.com/shinnn/spdx-license-ids/compare/v3.0.0...master


But the ^3.0.0 on spdx-expression-parse should be bringing this in.

storrisi:
There are others licenses that are failing in the same way, maybe an update to the spdx parse library would fix the problem

storrisi:
Looks like the versions of the involved libraries are already the latest available

dabutvin:
superceded by #472

PR #456, opened 2019-02-06
Suggestions rendering fixes

storrisi:
Now that the Suggestions are in place, i was able to test the behavior on the UI, and i found some issues with the ModalEditor that i fixed in this PR.

To test how the Suggestions works, you can check https://dev.clearlydefined.io//definitions/nuget/nuget/-/Newtonsoft.Json/10.0.2

### DIFF
--- a/src/components/Navigation/Sections/LicensedSections.js
+++ b/src/components/Navigation/Sections/LicensedSections.js
@@ -30,9 +30,7 @@ class LicensedSection extends Component {
       handleRevert,
       applyCurationSuggestion
     } = this.props
-    console.log(JSON.stringify(rawDefinition))
     const definition = Contribution.foldFacets(rawDefinition, activeFacets)
-    console.log(JSON.stringify(definition))
     const { licensed } = definition
     const totalFiles = get(licensed, 'files')
     const unlicensed = get(licensed, 'discovered.unknown')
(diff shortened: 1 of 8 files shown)


src/components/Navigation/Ui/EditableFieldRenderer.js
```diff
@@ -36,37 +34,34 @@ class EditableFieldRenderer extends Component {
   }
 
   componentDidMount() {
-    const { definition, previewDefinition, curationSuggestions, field, type } = this.props
-    this.setComputedValue(definition, previewDefinition, curationSuggestions, field, type)
+    const { definition, previewDefinition, field, type } = this.props
+    this.setComputedValue(definition, previewDefinition, field, type)
```
storrisi: In this file i've removed some old implementations of the first attempt for the Suggestions behavior

dabutvin:
how do you get to the suggestions in the UI?

src/components/Navigation/Sections/LicensedSections.js
```diff
@@ -30,9 +30,7 @@ class LicensedSection extends Component {
       handleRevert,
       applyCurationSuggestion
     } = this.props
-    console.log(JSON.stringify(rawDefinition))
     const definition = Contribution.foldFacets(rawDefinition, activeFacets)
-    console.log(JSON.stringify(definition))
```
dabutvin: 👍

storrisi:
@dabutvin when looking at the details of a definition, if a field is missing (actually only License and Source Url), and some suggested data are returned from the server, an icon shows up near to the field.
Once clicked, a list of the suggested data appears, so the user can choose the value to apply.

schermata 2019-02-06 alle 19 33 01

PR #460, opened 2019-02-07
Browse page improvements

storrisi:
Now the page calls the endpoint with filters attached to the query.
`continuationToken` has been added to manage pagination as well.
FilterBar has been improved in order to accept custom values for each of the shown filters.

### DIFF
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -71,7 +71,7 @@ class PageBrowse extends SystemManagedList {
     const target = event.target
     const activeProvider = target.name
     this.setState({ ...this.state, activeProvider })
-    this.props.dispatch(uiBrowseGet(this.props.token, activeProvider))
+    this.props.dispatch(uiBrowseGet(this.props.token, { type: activeProvider }))
   }
 
   render() {
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -16,7 +16,7 @@ import FilterBar from '../../Sections/FilterBar'
 import FullDetailPage from '../../../FullDetailView/FullDetailPage'
 import ProviderButtons from '../../Ui/ProviderButtons'
 import ContributePrompt from '../../../ContributePrompt'
-import { licenses, releaseDates } from '../../../../utils/utils'
+import { licenses } from '../../../../utils/utils'
 
 /**
  * Page that show to the user a list of interesting definitions to curate
@@ -87,6 +87,8 @@ class PageBrowse extends SystemManagedList {
         onSort={this.onSort}
         hasComponents={!this.hasComponents()}
         showSourceFilter={false}
+        showReleaseDateFilter={false}
+        showCurateFilter={true}
         customLicenses={licenses.filter(license => license.value !== 'absence' && license.value !== 'presence')}
       />
     )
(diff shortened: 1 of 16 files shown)


src/components/Navigation/Sections/FilterBar.js
```diff
@@ -10,45 +10,83 @@ export default class FilterBar extends Component {
     activeFilters: PropTypes.object,
     onFilter: PropTypes.func,
     onSort: PropTypes.func,
-    hasComponents: PropTypes.bool
+    hasComponents: PropTypes.bool,
+    showSortFilter: PropTypes.bool,
+    showLicenseFilter: PropTypes.bool,
+    showSourceFilter: PropTypes.bool,
+    showReleaseDateFilter: PropTypes.bool,
+    customLicenses: PropTypes.object,
```
dabutvin: should this be `customLicenses: PropTypes:Array` ?

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
```diff
@@ -63,15 +86,47 @@ class PageBrowse extends SystemManagedList {
...
-    this.props.dispatch(uiBrowseGet(this.props.token, { type: activeProvider }))
+    this.setState({ ...this.state, activeProvider, activeFilters: null, activeSort: null })
+    this.props.dispatch(
+      uiBrowseGet(this.props.token, { type: activeProvider, maxLicensedScore: 70, maxDescribedScore: 70 })
+    )
+  }
+
+  async updateData(continuationToken) {
+    const { activeFilters, activeSort, activeProvider } = this.state
+    const query = { type: activeProvider, maxLicensedScore: 70, maxDescribedScore: 70 }
```
dabutvin: why set `maxDescribedScore` and `maxLicensedScore` to 70?

I think we should only be setting filters if they are picked by the user

storrisi: The initial request for the Browse page was to show only definitions with a score below 70

src/components/Navigation/Sections/FilterBar.js
storrisi: Right, i will fix it, thanks!

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
```diff
@@ -15,15 +16,19 @@ import FilterBar from '../../Sections/FilterBar'
 import FullDetailPage from '../../../FullDetailView/FullDetailPage'
 import ProviderButtons from '../../Ui/ProviderButtons'
```
dabutvin: these provider buttons are mixing types and provider.

```jsx
    <ButtonGroup>
        <Button name="github" onClick={onClick} active={activeProvider === 'github'}>
          GitHub
        </Button>
        <Button name="maven" onClick={onClick} active={activeProvider === 'maven'}>
          Maven
        </Button>
        <Button name="npm" onClick={onClick} active={activeProvider === 'npm'}>
          NPM
        </Button>
        <Button name="nuget" onClick={onClick} active={activeProvider === 'nuget'}>
          NuGet
        </Button>
        <Button name="crate" onClick={onClick} active={activeProvider === 'crate'}>
          Crate
        </Button>
        <Button name="pypi" onClick={onClick} active={activeProvider === 'pypi'}>
          PyPi
        </Button>
        <Button name="rubygems" onClick={onClick} active={activeProvider === 'rubygems'}>
          RubyGems
        </Button>
      </ButtonGroup>
```

the providers are:
 - cratesio
 - rubygems
 - github
 - mavencentral
 - npmjs
 - nuget
 - cocoapods
 - pypi

the types are:
 - crate
 - gem
 - git
 - maven
 - npm
 - nuget
 - pod
 - pypi
 - sourcearchive

I think the best case would actually be to align the providerButtons with the actual list of providers here and then change the query from type to provider.
Make sense?

dabutvin:
when the list loads 50 entries, right now it makes a request to get the full definition for each one.
We need to render what we get back from the initial query and not call out for more data until a specific definition magnifying glass is picked.

The results from the query should be enough to render the TwolineEntry, the only thing missing is are the files

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
dabutvin: i see.

dabutvin: I just had a quick chat with @fossygirl on this.
Let's remove the <70 filter from the query by default and add a new 'curate' dropdown to turn on these filters

Here's a mockup of the experience

image

When you check 'licensed' filter by maxLicensedScore: 70
When you check 'described' filter by maxDescribedScore: 70

storrisi: Got it, i will implement this, thank you!

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
storrisi: These buttons are also used in the Harvest section, i need to check if this change would have some impact to that section

storrisi:
Thanks @dabutvin , i will implement:
- Curate filter
- Remove full definition detail for each definition in the list

storrisi:
@dabutvin i've updated this PR with these fixes:

* Added Curate filter to FilterBar
* Retrieve component details from data returned by the query
* Updated tests
* VsCode Chrome debug script
* Separation of browse and definition store

Talking about the last point, while testing the application i've noticed that the Redux store used for the Browse Page was the same used for the Definitions Page.
This ended up on have the components list populated while switching from the Browse to the Definitions page and viceversa.
Due to this behavior i've created another portion of the store specifically for the `definitions`, since the original was already called `browse`.

dabutvin:
👍 👍 👍

PR #464, opened 2019-02-11
Update link for github components

### LINKED ISSUE #463: update link for github components to open repo at commit
Currently clicking on the name of a github component in a definition list entry opens the commit associated with the revision. Something like `https://github.com/intel/tinycbor/commit/d94ca09aa91f5b3c581527aa8bca179a82b79874`

This opens the actual commit with may contain just one file or something. Rather, it would be more useful to open the *tree* like this `https://github.com/intel/tinycbor/tree/d94ca09aa91f5b3c581527aa8bca179a82b79874` so the user can directly browse around the project at that point in time. 

Make sure this is true for everywhere that we do this sort of thing.

storrisi:
Fixes #463

### DIFF
--- a/src/utils/entitySpec.js
+++ b/src/utils/entitySpec.js
@@ -139,4 +139,49 @@ export default class EntitySpec {
   toString() {
     return this.toPath()
   }
+
+  static getComponentUrl(coordinates) {
+    switch (coordinates.provider) {
+      case 'github':
+        return `https://github.com/${coordinates.namespace}/${coordinates.name}`
+      case 'npmjs':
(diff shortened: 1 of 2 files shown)


src/utils/entitySpec.js
```diff
@@ -139,4 +139,49 @@ export default class EntitySpec {
   toString() {
     return this.toPath()
   }
+
+  static getComponentUrl(coordinates) {
```
storrisi: I've moved these methods into a utils class to keep the original method more readable and to be able to re-use them somewhere else

PR #466, opened 2019-02-12
update code for browse page

dabutvin:
- rename to browseDefinitions api call to searchDefinitions to be reused
- use isFetching redux pattern for the spinner on browse page
- remove the spinner and threshold override from the infinite scroll
- tighten up the browse window
- remove storeList switches and use the inheritance to change behavior

### DIFF
--- a/src/actions/definitionActions.js
+++ b/src/actions/definitionActions.js
@@ -99,7 +99,7 @@ export function browseDefinitionsAction(token, query, name) {
     const actions = asyncActions(name)
     dispatch(actions.start())
     try {
-      if (!query.continuationToken) dispatch(uiBrowseUpdateList({ startQuery: true }))
+      dispatch(uiBrowseUpdateList({ startQuery: true }))
       const result = await searchDefinitions(token, query)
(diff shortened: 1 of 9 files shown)


dabutvin:
made some updates.

The preview is here: https://deploy-preview-466--clearlydefined.netlify.com/browse

storrisi:
The implementation looks fine.
I've checked the deploy and i suggest to add a spinner on each pagination request, since when the user reach the end of the table it take some seconds to see the results.
Thoughts?

dabutvin:
thanks @storrisi 
yeah I was going back and forth with that spinner.

This change looks if there is a continuationToken and does not show the spinner if we are continuing.
I'm happy to switch it back. @jeffmcaffer you have any opinions?

jeffmcaffer:
No strong opinion on my part at this time. Would rather see us doing more prefetching so the user doesn't need to wait...

dabutvin:
I played around with different thresholds a bit without much success. I think I'll add the spinner back until we figure this out

PR #467, opened 2019-02-13
License picker ui updates

dabutvin:
This change will let the user use the inline editor for editing licenses, but launch the license picker if they want to use the expression builder.


fossygirl:
The curators thank you! 👍

dabutvin:
Simplest case:




More complicated:

storrisi:
Thanks @dabutvin , once merged i will update e2e tests based on this new approach

dabutvin:
thanks @storrisi I'll look out for that update, once we get that branch merged I'll make sure to keep it up to date with my changes too 👍 

@jeffmcaffer @fossygirl feeling good about this ux with the eyeball icon to launch the expression builder?

dabutvin:
let's do it!

PR #471, opened 2019-02-15
Issue #470 - Revert action doesn't work on Definitions List

### LINKED ISSUE #470: Revert action doesn't work on Definitions list
Once updated the value of any field of a component in the Definitions list, the Revert button of the specific field doesn't restore the original value.

storrisi:
Fixes #470


src/components/LicensesRenderer.js
```diff
@@ -38,7 +38,7 @@ class LicensesRenderer extends Component {
     const { advancedView } = this.state
 
     return (
-      <div class="license-renderer">
+      <div className="license-renderer">
```
dabutvin: 👍

PR #472, opened 2019-02-15
drop spdx-expression-validate and just do try/catch (fixes #412)

### LINKED ISSUE #412: license picker gets in odd state

dabutvin:


storrisi:
This is a good workaround, but given the author of the lib just released an update, should we try to use it and see if it fix the behavior?

See: https://github.com/jslicense/spdx-expression-validate.js/commit/4a26b1373da4a461a15d2ca94810661ae719bf43

dabutvin:
right, but I think we need to leave it this way for our special handling of 'NONE'/'NOASSERTION'

does that make sense?

PR #477, opened 2019-02-21
E2E: Added back timeouts to test suites

storrisi:
This fixes a problem when running E2E locally.

Apparently the jest.setTimeout() doesn't work as expected.
Adding back the timeouts to each single test do the trick.

### DIFF
--- a/e2e/tests/definitions.js
+++ b/e2e/tests/definitions.js
@@ -50,7 +50,7 @@ describe('Definitions page', () => {
     await expect(page).toMatchElement(definitionsMap.component.switchButton)
     await expect(page).toMatchElement(definitionsMap.component.revertButton)
     await expect(page).toMatchElement(definitionsMap.component.removeButton)
-  })
+  }, 20000)
 
   test('should display the detail after clicking on a component in the list', async () => {
     await page.waitForSelector(definitionsMap.component.firstElement)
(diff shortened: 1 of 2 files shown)


dabutvin:
you may have to `npm i` to make sure you have the latest version of pupeteer

unfortunately this is not long enough for the crappy CI boxes we are using
can we try adding a default instead of adding to each individual?

```
beforeAll(async () => {
    jest.setTimeout(process.env.JEST_TIMEOUT || 20000)
    browser = await puppeteer.launch({ headless: process.env.NODE_ENV !== 'debug', slowMo: 80 })
    page = await browser.newPage()
    page.setDefaultTimeout(process.env.JEST_TIMEOUT || 20000)
    await page.setViewport({ width: 1920, height: 1080 })
    await page.goto(`${__HOST__}/definitions`, { waitUntil: 'domcontentloaded' })
  })
```

storrisi:
I already have the latest version, but increasing the default timeout doesn't work on the local environment. 
I will search for a solution.

storrisi:
@dabutvin take a look on our latest solution, we've set a defaultTimeout value, that could be also modified during the CI process.
It should be easy enough to be maintained time by time.

What do you think?

e2e/tests/definitions.js
```diff
@@ -1,143 +1,150 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 import { definitionsMap } from '../maps/definitions'
+import { setDefaultOptions } from 'expect-puppeteer'
+
 const puppeteer = require('puppeteer')
+const defaultTimeout = process.env.JEST_TIMEOUT ? process.env.JEST_TIMEOUT : 30000
```
storrisi: This value needs to be set in these 3 points:
- puppeteer configuration
- jest configuration
- test callback

Otherwise, the entire test will fail.

dabutvin:
👍 👍 lgtm

PR #483, opened 2019-02-22
Fix issue #212 - Navigation header items disappear

### LINKED ISSUE #212: Navigation header items disappear when view width is too narrow.
1. Load clearlydefined.io
2. Drag window edge so that the window becomes more narrow
3. Notice that the navigation items at the top disappear

storrisi:
Fixes #212


dabutvin:
Nice! I like the approach of breaking up the layout work a bit

Can we space these menu items evenly?

image

I'm thinking we should reduce the space between 'About' and 'Login' and 'Login' and 'Docs' to match the space between 'Definitions' and 'About'

gianpaj:
Thanks for catching that @dabutvin. I've fixed that

src/components/Header.js
```diff
@@ -91,18 +85,21 @@ class Header extends Component {
...
         <Navbar.Header>
           <Navbar.Brand>
             <LinkContainer to={ROUTE_ROOT}>
               <img src={logo} className="App-logo" alt="logo" />
             </LinkContainer>
           </Navbar.Brand>
+          <Navbar.Toggle />
         </Navbar.Header>
         <Navbar.Collapse>
           {this.renderNavigation(navigation, session.isAnonymous)}
```
dabutvin: when these are clicked, can we close the menu?
 
Right now, I am seeing it stay open

gianpaj:
Done. It was a simple prop 🙂

PR #484, opened 2019-02-22
Address layout of component list panel #74

### LINKED ISSUE #74: Address layout of component list panel
Narrow your browser a bit:

image

At first glance it seems like the infinite list stuff seems to be adding fixed heights and using absolute positioning which doesn't play well with bootstrap's responsive css.

storrisi:
Fixes #74 

We've changed a little bit the layout of the Components list in the mobile view.
When the deploy will be ready, you should be able to watch how it looks.


dabutvin:
when opening a definition using the magnifying glass the modal is all the way off screen, can we have it visible at the top? The modal frame should also not have horizontal scroll

storrisi:
Thanks for the feedback @dabutvin we'll address this fixes soon

jeffmcaffer:
Nice changes. Could address the footer here or in another PR as well?

storrisi:
@jeffmcaffer yes, we're currently working on that and we'll open a separate PR.

src/components/Navigation/Ui/ComponentButtons.js
```diff
@@ -58,88 +59,105 @@ export default class ComponentButtons extends Component {
...
-                onClick={() => this.revertComponent(component)}
-                disabled={!hasChange(component)}
-              >
-                <i className="fas fa-undo" />
-              </Button>
-            </ButtonWithTooltip>
-          )}
-        </ButtonGroup>
+        {isMobile ? (
+          <ButtonToolbar>
```
jeffmcaffer: suggest breaking this out into a separate function as well.

src/components/Section.js
```diff
@@ -5,25 +5,44 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { Row, Col } from 'react-bootstrap'
```
jeffmcaffer: extra blank line

src/styles/_App.scss
```diff
@@ -101,6 +101,14 @@
   margin: 30px;
 }
 
+.spacer-search {
+  margin: 3rem;
+
+  @media screen and (max-width: 992px) {
```
jeffmcaffer: Feels like we should have constants for these various max-width values.

BTW, why 991/2? is that a standard somehow. My phone (Samsung S8+) is 768 in portrait mode

src/styles/_List.scss
```diff
@@ -57,11 +57,11 @@
 .list-letter {
   display: inline-block;
   height: 40px;
-  width: 40px;
+  min-width: 40px;
   margin-right: 20px;
   line-height: 40px;
   text-align: center;
-  border-radius: 40px;
+  border-radius: 50%;
```
jeffmcaffer: are the letter circles actually smaller (or larger for that matter) on mobile? How does that jive with the icons when we have them?

src/utils/utils.js
```diff
@@ -66,4 +66,20 @@ const releaseDates = [{ value: 'presence', label: 'Presence Of' }, { value: 'abs
 
 const curateFilters = [{ value: 'licensed', label: 'Licensed' }, { value: 'described', label: 'Described' }]
 
-export { setIfValue, difference, customLicenseIds, asObject, sorts, licenses, sources, releaseDates, curateFilters }
+const isMobile = window.innerWidth < 991
```
jeffmcaffer: the reference to "mobile" here feels overly specific. It is really "narrow". Could be narrow for lots of reasons, or indeed, wide on a mobile device.

dabutvin: we could stick with the bootstrap 3 terminology

```
/* Extra small devices (phones, less than 768px) */
/* No media query since this is the default in Bootstrap */

/* Small devices (tablets, 768px and up) */
@media (min-width: @screen-sm-min) { ... }

/* Medium devices (desktops, 992px and up) */
@media (min-width: @screen-md-min) { ... }

/* Large devices (large desktops, 1200px and up) */
@media (min-width: @screen-lg-min) { ... }
```

from https://getbootstrap.com/docs/3.4/css/

jeffmcaffer: That sounds good. are we going to handle the multiple sizes or just <= small and >= medium? Recall my comment elsewhere here that my phone (recent-ish and clearly mobile) is 768 wide. So if we are going to do phones, feels like we have to target 768.

src/utils/withSuggestions.js
```diff
@@ -24,10 +24,10 @@ function withSuggestions(WrappedComponent) {
     render() {
       const { suggestedData, field } = this.props
       return (
-        <div style={{ display: 'flex', alignItems: 'center' }}>
+        <>
```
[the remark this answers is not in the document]
gianpaj: it's the short syntax for a [React.Fragment](https://reactjs.org/docs/fragments.html). It doesn't render any DOM element. It's used here because the render method needs to return a single component

src/styles/_List.scss
gianpaj: I can leave the 40px fixed size. It's just a % works for any size element.
Should leave the size in pixels?

src/utils/utils.js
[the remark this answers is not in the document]
gianpaj: That's true. This is was not listening for screen size changes. I assumed that was not necessary since, for example, the iPhone X height is 812px.

But I've made the change now

gianpaj: Sounds good. I've fixed the CSS media queries

src/components/FullDetailView/AbstractFullDetailsView.js
```diff
@@ -33,7 +35,8 @@ export class AbstractFullDetailsView extends Component {
       <Modal
         closable={false}
         footer={null}
-        centered
+        // if it's mobile do not center the Modal
+        centered={!isMobile}
```
dabutvin: I'm seeing the modal open halfway down the page on mobile width.
Is there anything we can do to align it to the top?

src/components/FullDetailView/AbstractFullDetailsView.js
```diff
@@ -33,7 +35,8 @@ export class AbstractFullDetailsView extends Component {
...
         destroyOnClose={true}
         visible={visible}
         width={'85%'}
```
dabutvin: the modal has a horizontal scroll at mobile width. Is there something overriding this?

gianpaj:
Hi @dabutvin,

I've fixed the issue. I used the HOC in the wrong place.

FYI the modal fixes of the Full Detail view are in https://github.com/clearlydefined/website/pull/510

gianpaj:
Hi @dabutvin ,
I've updated the branch with the latest changes from master.

Can you please take another look?

src/utils/WindowProvider.js
```diff
@@ -0,0 +1,64 @@
+/* Copyright (c) Microsoft Corporation and others. */
```
jeffmcaffer: // comments for license and copyright

src/utils/WindowProvider.js
```diff
@@ -0,0 +1,64 @@
...
+/* SPDX-License-Identifier: MIT */
+
+import React, { PureComponent, createContext } from 'react'
```
jeffmcaffer: BTW, did this code come from somewhere or did you write it from scratch?  Seems quite different. uses `var` not `const`, `function` rather than `=>`, names like `obj`, `func` etc that don't fit in the overall coding style.

src/styles/_List.scss
jeffmcaffer: I'm mostly interested in whether or not the size is changing. For example, the change from `width` to `min-width` and making the radius proportional might indicate that in some cases the width would be larger than 40px. Is that the intention? 

Related, should the change to use `min-width` be accompanied by a change in the use of `height`?

In the end I'm not really sure why these changes are needed or the intent

src/utils/WindowProvider.js
gianpaj: I took the `throttle()` function from MDN:
https://developer.mozilla.org/en-US/docs/Web/Events/resize#requestAnimationFrame_customEvent
I'll update the syntax

src/styles/_List.scss
gianpaj: > I'm mostly interested in whether or not the size is changing. For example, the change from `width` to `min-width` and making the radius proportional might indicate that in some cases the width would be larger than 40px. Is that the intention?

If we set it to `width: 40px ` the icon will be smaller on mobile size screens. Like this:



> Related, should the change to use `min-width` be accompanied by a change in the use of `height`?

Looks like the height of this `<span>` is determined by the `line-height: 40px;`
But the `height` style limits its height so it doesn't grow over the 40px:

mar-06-2019 20-56-02


> In the end I'm not really sure why these changes are needed or the intent

It was merely to fix the Maven icon on Mobile size screens so it doesn't look like it has shirked

gianpaj:
After merging from master #515, the component list looks like this on mobile:



How should we display if there are "Pending curations"?

Maybe it will fit if the Score badge when it be displayed as a single number? (see #545)

For example, we can make the font size of "Pending curations" badge smaller:



_in mockup I used `font-size: x-small;`_

jeffmcaffer:
Dropping the font size might be good. Could also drop words. E.g., drop "curations" and have just "Pending". Should definitely do these changes in the context of the new buttons and badges work in #545

jeffmcaffer:
Looks good though I was unclear. I meant to change "Pending curations" to "Pending" when the screen is small, not always.

jeffmcaffer:
There have been many many changes since this was posted. Likely the problem and the code need to be revisited.

gianpaj:
Hi @jeffmcaffer I've fixed all the UI and tests issues.

storrisi:
@jeffmcaffer @dabutvin any chance you could review and hopefully merge this one?

src/utils/WindowProvider.js
dabutvin: instead of taking the throttle from MDN, can we use lodash throttle?
They have updated docs for using a timeout throttle instead of requestAnimationFrame

https://developer.mozilla.org/en-US/docs/Web/Events/resize#setTimeout

I think it would look something like this:

```js
  componentDidMount() {
    window.addEventListener('resize', _.throttle(this.updateDimensions))
  }

  componentWillUnmount() {
    window.removeEventListener('resize')
  }
```

what do you think?

gianpaj: sure. sounds good

gianpaj: I've used lodash and use debounce since it makes more sense than throttle in this case

storrisi:
@jeffmcaffer @dabutvin is there something that is blocking this PR to be merged? In that case, we'll be happy to fix any outstanding problem.

dabutvin:
thanks @storrisi @gianpaj this is great!

Issue #343, opened 2018-10-24
Declaring license under "Files" view of component


bduranc:
Greetings,

As current, when I go to declare a license for a file under the source tree of a component ("File" view), the source tree refreshes and collapses every time I declare a license at that level.

In other words, if want to declare licenses for multiple files under that source tree, I need to expand the tree again and find that file.

![stepscombined](https://user-images.githubusercontent.com/31491646/47451111-c7edcd80-d77b-11e8-8b1f-1a5c39e44590.png)

This is really time consuming as essentially the above steps need to be done for every file I declare a license for.

What I think would be more efficient, is the ability to declare licenses for multiple files in the source tree without having the tree refresh on the user and collapse.

AlexWebYourmind:
@bduranc this is a known issue, related to react-table. We have opened an issue on their repository, that is currently being addressed as part of future implementations --> https://github.com/react-tools/react-table/issues/1062

Should it not be fixed in a reasonable time, we'll probably need to evaluate alternative implementations.

@jeffmcaffer @storrisi thoughts?

storrisi:
@jeffmcaffer we'll look further into this. 
From our past investigations, we ended up that it could be easier to switch to [antd/table](https://ant.design/components/table/#components-table-demo-expand-children) , which is the matter we brought ant-design as a library into the project.

jeffmcaffer:
There are a number of these sorts of things that should be addressed soon. Perhaps for January we should plan a UI pass to fix a bunch of the nits and pain points. @storrisi thoughts?

storrisi:
I agree, we could plan for January an activity to fix these kind of issues.

storrisi:
This one is related to #489 , probably the automation was wrong. It could be closed.

PR #489, opened 2019-02-25
FileList implementation with ant-design

### LINKED ISSUE #469: FileList collapses when fields are edited
The tree should stay in it's current state when a change is made. Right now, when a change is made it collapses and has to be reopened each time something changes

storrisi:
According on what discussed on #469 , we've used another library to render the FileList.
This implementation switches from the old `react-table` to the Table element available in `ant-design`

Fixes #469 , #343


jeffmcaffer:
Have not looked at the code but the UX looks pretty good! How styleable is it? For example, would like to 
* remove the line between files
* make each row shorter (smaller height)
* change the amount of indentation on the tree part (need to deal with long paths in Java)

How do you feel about the code?

storrisi:
The library has a lot of props, so i think we'll be able to apply most of the customizations.

Otherwise, we could use a classic style approach where a specific props doesn't exists.

See https://ant.design/components/table/#API

storrisi:
> How do you feel about the code?

A lot less complicated than `react-table`.
Many of the functionalities that we've implemented in a custom way on `react-table` , here are ready out of the box.
This seems very powerful and easy to use.

storrisi:
I've fixed the filter methods, this is now ready to be reviewed.

storrisi:
Don't know why E2E are failing. I haven't made any change on them since my last commit.

src/components/LicensesRenderer.js
```diff
@@ -53,7 +53,7 @@ class LicensesRenderer extends Component {
           onRevert={onRevert}
           type="license"
         />
-        <i className="fas fa-eye license-advanced" onClick={this.toggleAdvancedView} />
+        {!readOnly && <i className="fas fa-eye license-advanced" onClick={this.toggleAdvancedView} />}
```
dabutvin: 👍

src/components/FileList/FileList.js
```diff
@@ -1,275 +1,215 @@
...
- */
 export default class FileList extends Component {
-  static propTypes = {
-    readOnly: PropTypes.bool
+  state = {
+    files: [],
+    filteredFiles: [],
+    filteredInfo: {},
+    sortedInfo: {},
+    expandedRows: [],
```
dabutvin: what do you think about expanding the first row by default?

storrisi: It depends from the list of files.
When the list is composed only by a folder, then it make sense, otherwise the user could see a folder already opened that is not supposed to be.

dabutvin: ahh true - that makes sense. just a thought I had

PR #490, opened 2019-02-25
Issue #486 - Fix Mobile view of Footer and Homepage

### LINKED ISSUE #486: Fix mobile view of Homepage and Footer
schermata 2019-02-25 alle 11 30 59

Related to #95

storrisi:
Fixes #486


src/components/Footer.js
```diff
@@ -29,14 +29,14 @@ export default class Footer extends Component {
...
         </Col>
-        <Col sm={4} className="Footer-element Footer-center">
+        <Col xs={4} className="text-center">
           Are you ClearlyDefined?
         </Col>
-        <Col sm={4} className="Footer-element">
-          <div className="pull-right Footer-right">
+        <Col sm={4} xs={4} xsOffset={1}>
+          <div className="pull-right right">
             <a href="https://docs.clearlydefined.io/legal/terms" target="_blank" rel="noopener noreferrer">
```
dabutvin: can we force these to not wrap?

original:
image

proposed:
image

src/components/PageLanding.js
```diff
@@ -106,7 +106,7 @@ class PageLanding extends Component {
           </Col>
         </Row>
         <Row className="show-grid neighborhood-row">
-          <Col md={textColWidth}>
+          <Col xs={textColWidth}>
```
dabutvin: can we push/pull these columns so the icons all line up on the same side for small screens? I think the desktop view can stay as existing.
https://getbootstrap.com/docs/3.4/css/#grid-column-ordering

original:

image

proposed:
image

src/components/Footer.js
```diff
@@ -29,14 +29,14 @@ export default class Footer extends Component {
     const { socials } = this.props
     return (
       <footer className="Footer">
-        <Col sm={4} className="Footer-element">
```
dabutvin: can we add a layer of `<div class="container">` here so it doesn't go full browser width?

src/components/Footer.js
gianpaj: looks good. Should I hide the "pipe" characters "I" only for mobile?

gianpaj:
Hi @dabutvin, I've done this as well

dabutvin:
👍 nice job

PR #492, opened 2019-02-25
Issue #476 - Fix Facets background color when edited

### LINKED ISSUE #476: Modify curation review view to show facet changes
Right now the curation review view (e.g. https://clearlydefined.io/curations/1078) lights up the change the curator has made, for example a change to the definition, however it's very simple and not detailed enough to show facet editing.

We need to expand out this view to allow reviewers to see changes to facets easily. This page might need to be an augmented Full Details Page, perhaps, with highlighting of the facets a curator changed. We need to give the reviewers a much more at-a-glance understanding of what files will be ending up in the new facets if the change is accepted.

storrisi:
Fixes #476


dabutvin:
👍

PR #493, opened 2019-02-25
Issue #482 - Add guidance on popup for curation submission

### LINKED ISSUE #482: Add guidance on popup for curation submission
Since we changed the generation of the PR title to be the title the user enters on their curation submission popup, we should tell them in the help text that "this text will be used to generate the title of the PR." Might make them slightly more likely to enter something meaningful. :-)

@dabutvin for visibility

storrisi:
Fixes #482

### DIFF
--- a/src/components/ContributePrompt.js
+++ b/src/components/ContributePrompt.js
@@ -60,8 +60,32 @@ export default class ContributePrompt extends Component {
     return type !== 'select' && summary.length > 0 && details.length > 0 && resolution.length > 0
   }
 
+  renderTypeField() {
+    return (
+      <FieldGroup
+        className="inlineBlock pull-right"
+        name="type"
+        label="Type"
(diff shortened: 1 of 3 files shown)


gianpaj:
For quicker review here is the text in Summary field in the curation Popup Modal:
clearlydefined

jeffmcaffer:
Perhaps we should just change the UI to "Title" if it actually is going to be the title.

Then the placeholder can be something like "Short (100 char) description. Used as PR title"

gianpaj:
Sounds good. I've made those changes

gianpaj:
I've made the style changes as per #504.

@jeffmcaffer I've left the Type selector on the top right as it was originally suppose to look. There was missing CSS class.

Is this ok?
clearlydefined

dabutvin:
👍

PR #507, opened 2019-02-26
Browse: sort provider names

### LINKED ISSUE #494: Browse: sort provider names
The current list in the Providers drop down on the /browse page appears to be unsorted. Should be sorted alphabetically.

storrisi:
Fixes #494


dabutvin:
👍

PR #508, opened 2019-02-26
Multiselect for components

### LINKED ISSUE #169: Multiselect for components
The contribution UI should allow users to group and multiselect sets of components and contribute changes together as once. They should be able to review the list of changed definitions in the contribution and provide a summary descriptions for that contribution. Users should also get an option for "select all" or "unselect all."

storrisi:
Fixes #169


gianpaj:
GIF:
multiselect

jeffmcaffer:
looks good. lets go with that. We can also use the notion of selected in when contributing. That is, pick 7 of 9 modified things and contribute in one PR. then pick the remaining 2 and contribute a separate PR.

dabutvin:
👍 we should absolutely only include the selected components when creating a contribution. Good call

gianpaj:
I've noticed that if you update a license when the component is collapsed you might not see immediately that any change has been made (only the icon changes color).

multiselect-license

What suggestion would you have to make this more apparent?
Few ideas:
1. Automatically expand all the components that were changed to see those changes
2. Highlight the license has been changed (the one under the name of the component). The limitation with this option is that the Source and Release Date can be changed but are not currently show when collapsed

gianpaj:
> looks good. lets go with that. We can also use the notion of selected in when contributing. That is, pick 7 of 9 modified things and contribute in one PR. then pick the remaining 2 and contribute a separate PR.

Does a user have to select a component(s) to make a Contribution?
or, for example, if no components are selected, all components will be Contributed (as it's done now)

Which one does make more sense @jeffmcaffer , @dabutvin ?

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
```diff
@@ -217,6 +217,7 @@ class PageBrowse extends SystemManagedList {
           <div className={classNames('section-body flex-grow', { loading: components.isFetching })}>
             <i className="fas fa-spinner fa-spin" />
             <ComponentList
+              multiSelectEnabled={false}
```
storrisi: I think this property should be moved to the SystemManagedList or UserManagedList components.

If it's a System Managed list, the `multiSelectEnabled` should be false by default.
Instead, if we are in a User Managed List, that it should be true.

src/components/Navigation/Pages/PageDefinitions/PageDefinitions.js
```diff
@@ -97,11 +97,51 @@ export class PageDefinitions extends UserManagedList {
     )
   }
 
+  toggleCheckbox = index => {
```
storrisi: I suggest to` move toggleCheckbox , onSelectAll, onFieldChange,  resetComponents`,  methods to the User Managed List component.

src/components/Navigation/Sections/FilterBar.js
```diff
@@ -1,33 +1,148 @@
...
+    release: null,
+    source: null
+  }
+
+  onFieldChange = (field, value) => {
+    const { onFieldChange } = this.props
+    this.setState(prevState => {
+      // toggle
+      if (prevState[field] == value) {
+        onFieldChange(field, null)
```
storrisi: A check if this `onFieldChange`  method is not defined should be added

src/utils/withSuggestions.js
```diff
@@ -25,10 +25,10 @@ function withSuggestions(WrappedComponent) {
     render() {
       const { suggestedData, field } = this.props
       return (
-        <div style={{ display: 'flex', alignItems: 'center' }}>
```
storrisi: Are you sure about removing this style here?
How the suggestion icon would be renderer without this?

gianpaj: you're right. i'll fix this

jeffmcaffer:
IMO we should NOT auto expand all the selected items when a value is changed. That would be quite disruptive. Some suggestions 

* The list should have a count something like `N of M selected` sort of like GitHub does with 
image

* Report in a Notification how many definitions were changed
* Be able to filter to just show changed definitions
* in the Contribution dialog show the list of components with changes/being contributed

All of these can be implemented incrementally

gianpaj:
I've added the tests and implemented the required features

gianpaj:
I've fixed the conflicts and the tests are passing

bduranc:
> GIF:
> multiselect

From an end-users point of view, I will +1 this. 

Right now, using the work list is difficult because when you click Contribute, everything is submitted under one PR. This is OK for one or two components that have minor curation (say only top-level license data), but becomes really cluttered to file the single PR when doing > 2 components with more specific curation work, say being done on the file level.

I think having this multi-select ability would be a huge improvement to the existing contribution workflow if we could select only items we want to have in the contribution PR. This way curators could "stagger" their PRs out a bit to coincide with specific components and cases.

storrisi:
@dabutvin do you have the chance to review this one? I think it could be a great improvement

jeffmcaffer:
Looking at the UX, some suggestions

* The check boxes should be vertically centered in each row (including the header row) Looks misaligned with the type icon beside
* Would be good for there to always be a count of definitions in the list. E.g., 132 definitions. Then if there are some selected it can read 14/132 definitions or some such
* The Source, Release Date, License entries in the header row are a bit hard to distinguish as menus/clickable things. Looking at the GtiHub issues UI (and the right hand side of our UI), drop downs should have the down caret, perhaps the Source and Release ones could have '...' or some other decorator to hint that they are not just words. May also help to space them out a bit.
* I'm unsure about Source location being a bulk settable value. Very rare for two definitions to have the same source location.
* setting the license using the netlify deployment did not seem to work. I added a couple of MIT licensed definitions, selected both, and set the license but the change did not seem to appear anywhere in the UI. Note that it did work when i tried again later with some different components. Not sure...
* The select all check box (top left) appears all by itself when the workspace is empty. Looks off. If we have the count (e.g., 0 definitions) beside, that might look ok
* The changed license values are not showing up int he two line summary view. you can only see the changes if you expand the definition panel.Leads to some confusing UI where you see inconsistent licenses and see "old" values when you think you have changed them.
* note specifically as a result of this change but the Contribute dialog should show at least a count fo the definition changes being contributed. Ideally the dialog would show a list of the changed components but that's a topic for another PR,

storrisi:
Thanks @jeffmcaffer i'm taking care of your suggestions

storrisi:
I've made some changes:

- show always the number of definitions, this will also fix #595 
- vertical alignment of checkbox
- show the edit and revert icon on Source and Release date fields
- show the border around the License Picker
- show a different background on the license in the title if changed

jeffmcaffer:
Thanks @storrisi. Looking better. I do still wonder whether the source location and release date are reasonable things to have in the bulk/multi-select workflow.

We likely should also space thing out a bit more. looks pretty cluttered. May have to trim some words etc. Of course, if we remove the source location and release date entries then we win a bunch of space :-)

storrisi:
@jeffmcaffer i'm happy with your solution, so i've removed the Source and Release Date pickers.
We can even re-add those elements back in the future if we think we need them.

dabutvin:
Thanks for working through this. 

In an effort to bring the filter/sort bar and the multiselect together, I'd like to try a different design pattern - moving the license picker in-line with the definition entries. 
So when you make any change to a selected component, that change is reflected to all selected components.

With this kind of change, we should open a modal to confirm/cancel the change with wording that says.
```
"This change will apply to all 5 selected components"
         
                                                                | OK | Cancel |

```

With the license picker out of the way, let's try to squeeze the select-all checkbox and the sort/filter bar on the same row there with the same style as the homepage.

I know this is kind of abstract, happy to chat more about any ideas

jeffmcaffer:
Suggest a tweak to your tweak. the OK / Cancel choice makes it hard for people who have a multiselection to then just change one value. Rather, this could be 

```
You have N definitions selected. Apply this change just here or to all selected definitions. 
                                                                 | Change all | Change just this one |
```

All wording subject to tweaks.

storrisi:
@dabutvin thanks for your suggestion.

So, just as a recap, are you suggesting to let the user select multiple components, and from there any change applied to a single definition could be replicated to the entire selection?

If that is correct, i agree with @jeffmcaffer about the modal message, and we should also consider that also any revert function could affect all the selected definitions.

I'm wondering if that could be clear from a user point of view, maybe once a selection is made we should show a message somewhere, like "Edit a single definition to apply the same change to the entire selection"

storrisi:
What about something like this?

Schermata 2019-04-19 alle 17 57 52

storrisi:
Here's a quick gif of the current behavior after the suggested changes (i have still to add the confirmation modal)

Screen Recording 2019-04-19 at 07 24 PM

jeffmcaffer:
I like the inline nature of this. Nice. I was thinking that with multiple selected, changing the value on one would then prompt the user something along the lines of 

```
You have N definitions selected. Apply this change to all or just this definition? 
                                             | Change all | Change just this one |
```

storrisi:
@jeffmcaffer i've added a notification every time a user makes a change to a selection of definitions
Schermata 2019-04-24 alle 17 49 42

storrisi:
I think we have almost completed this implementation, we can then improve it by adding the Filter / Sort exposure on the Workspace page on another PR. Do you agree?

PR #515, opened 2019-02-27
surface curations and contributions in the UI

dabutvin:
wire up the new api from https://github.com/clearlydefined/service/pull/412 to the existing ui


src/components/Navigation/Ui/ComponentButtons.js
```diff
@@ -59,14 +61,14 @@ export default class ComponentButtons extends Component {
...
+    const { definition, curations, currentComponent, readOnly, hasChange, hideVersionSelector } = this.props
     const component = EntitySpec.fromCoordinates(currentComponent)
     const isSourceComponent = this.isSourceComponent(component)
     const scores = Definition.computeScores(definition)
     const isDefinitionEmpty = Definition.isDefinitionEmpty(definition)
     const isSourceEmpty = Definition.isSourceEmpty(definition)
-    const isCurated = Definition.isCurated(definition)
-    const hasPendingCurations = Definition.hasPendingCurations(definition)
+    const isCurated = Curation.isCurated(curations)
+    const hasPendingCurations = Curation.hasPendingCurations(curations)
```
dabutvin: `curations` here is undefined, still needs to be wired up

src/actions/ui.js
```diff
@@ -86,8 +86,8 @@ export function uiNotificationDelete(message) {
   return { type: UI_NOTIFICATION_DELETE, message }
 }
 
-export function uiInspectGetCuration(token, entity) {
-  return getCurationAction(token, entity, UI_INSPECT_GET_CURATION)
+export function uiInspectGetCurations(token, entity) {
+  return getCurationAction(token, entity, UI_INSPECT_GET_CURATIONS)
```
storrisi: If you are passing `UI_INSPECT_GET_CURATIONS` has a third param , i suggest to change `export function getCurationAction(token, entity)` to `export function getCurationAction(token, entity, name)`at https://github.com/clearlydefined/website/pull/515/files#diff-5aa8d7146713f00ce77323a034bcc4f9R14
instead of import the action name.

src/components/Navigation/Ui/ComponentButtons.js
storrisi: To do that , `curations` should be returned by invoking the /curation API for each component in the list.
Is it really worthwhile or maybe we could consider to have these data in the /definition API response? 
Those are only needed to show the status of the curation (isCurated, hasPendingCurations).

dabutvin: I agree. not worthwile, let's back this out

src/actions/ui.js
dabutvin: makes sense 👍

dabutvin:
@storrisi can you take another pass on this one before we merge?

src/components/Navigation/Ui/ComponentButtons.js
```diff
@@ -65,13 +64,10 @@ export default class ComponentButtons extends Component {
     const scores = Definition.computeScores(definition)
     const isDefinitionEmpty = Definition.isDefinitionEmpty(definition)
     const isSourceEmpty = Definition.isSourceEmpty(definition)
-    const isCurated = Definition.isCurated(definition)
-    const hasPendingCurations = Definition.hasPendingCurations(definition)
+
     return (
       <div className="list-activity-area">
         {scores && <ScoreRenderer scores={scores} definition={definition} />}
-        {isCurated && <Tag color="green">Curated</Tag>}
```
storrisi: Removing this tags will cause the user to don't know what's the status of a definition in the list without inspecting it.
I think it could be useful for a user to know the status at this stage, but i could be wrong.

My past suggestion was not intended to remove these tags, but instead of take the curation status from the /curations API, which require use to make a new /curations call for each component in the list, we could receive those params as additional in the /definitions API, as already made for the urls property.

What do you think?

dabutvin: I see. I don't think we should add a status like pending curation to the definition response. This is too time dependent, and we should be able to cache the definition regardless if a PR is open or closed against it.

We could however check the definitions described/tools to see if there are any curations that are part of it. This would allow us to add the 'Curated' tag here. I'll push that update now

storrisi:
Looks good to me.
I had only a doubt about showing the curation status in the Definition list.

Unfortunately i'm unable to see how this implementation works since it needs to have the service part merged, to see the results over the deployed version.

dabutvin:
thanks @storrisi 

We will have a chance to review everything e2e in dev tomorrow before we release to prod

dabutvin:
update on this.

in the definitionList (componentButtons) we will only show 'pending curations' badge and 
make it a link to the PR pending

We will not show 'curated' badge in the definition list

storrisi:
This one fixes #515

dabutvin:
image
image

PR #520, opened 2019-02-27
Use definition urls from service

### LINKED ISSUE #462: Have option to download the actual component
There are times when reviewing a definition that it would be good to get the actual component on the local drive and look at the files. We cache a lot of the interesting files but not everything. We could add a `downloadUrl` property to `described` domain and then show it in the UI, or have the UI compute the download URL.  In theory we could also just add a GET API on the service that gets the URL but that seems like a lot of overhead for a relatively simple thing that we've already computed.

I'd lean towards doing it in the definition as that is more durable and enables multiple clients to get the same behavior without duplicating logic.

The download button (whatever form it takes) can just be on the Full Details page I think. perhaps near the Files section. Seems most relevant. Open to options...

storrisi:
Fixes #462 

I've added the Download button in the FullDetailsView, but i'm not so happy about the placement.

This is where the button shows right now:

schermata 2019-02-27 alle 22 36 59

Any feedback on this is appreciated

This is related to https://github.com/clearlydefined/service/pull/410 so it won't work until that PR will be merged on the service.


dabutvin:
maybe here?

image

storrisi:
Thanks, way much better

storrisi:
@dabutvin I've moved the button following your suggestion

src/components/ComponentList.js
```diff
@@ -84,8 +84,8 @@ export default class ComponentList extends React.Component {
     } = this.props
     const component = list[index]
     if (!component) return
-    let definition = this.getDefinition(component)
-    definition = definition || { coordinates: component }
+    const definition = this.getDefinition(component) || { coordinates: component }
+    console.log(definition)
```
dabutvin: remove console log

src/components/DefinitionEntry.js
```diff
@@ -83,10 +87,10 @@ export default class DefinitionEntry extends React.Component {
         {name}
       </span>
     )
-    const revisionTag = revisionUrl ? (
+    const revisionTag = definition.described.urls.version ? (
```
dabutvin: should probably use `get(item, 'described.urls.version')` here to be compatible with previous

storrisi:
@dabutvin i've updated the PR and resolved some conflicts due to the recent merge on master branch

src/components/DefinitionEntry.js
```diff
@@ -83,10 +87,10 @@ export default class DefinitionEntry extends React.Component {
         {name}
       </span>
     )
-    const revisionTag = revisionUrl ? (
+    const revisionTag = get(definition, 'described.urls.version') ? (
       <span>
         &nbsp;&nbsp;&nbsp;
-        <a href={revisionUrl} target="_blank" rel="noopener noreferrer">
+        <a href={get(definition, 'described.urls.registry')} target="_blank" rel="noopener noreferrer">
```
dabutvin: typo. this should be 

```html
        <a href={get(definition, 'described.urls.version')} >
```

src/components/DefinitionEntry.js
```diff
@@ -325,6 +330,7 @@ export default class DefinitionEntry extends React.Component {
 
   render() {
     const { definition, onClick, renderButtons, component, draggable } = this.props
+    console.log(definition, component)
```
dabutvin: remove console log

storrisi:
There were some conflicts due to the recent changes on the definitions list. I've resolved those.

storrisi:
Yes, the service work is addressed here https://github.com/clearlydefined/service/pull/426 and it's ready as well.

PR #523, opened 2019-02-28
Fix Filelist e2e

storrisi:
Since the new Filelist has been merged, some e2e tests error showed up.

This one will fix those errors.


dabutvin:
:+1:

PR #524, opened 2019-02-28
Browse: align Revert, Collapse and Contribute buttons

### LINKED ISSUE #497: Browse: Align Revert, Collapse and Contribute buttons
On the /browse page the Revert, Collapse and Contribute buttons at the top right should be vertically aligned with the filter box. As it is I suspect it is being vertically centered in the div and the div became two lines high with the addition of the filter box.

See the /definitions page for proper alignment

storrisi:
Fixes #497


src/styles/_App.scss
```diff
@@ -244,3 +244,18 @@
...
+.section-header div:nth-child(2) {
+  align-self: flex-end;
+}
+
+@media (min-width: 768px) {
+  .section-button .text-right * {
+    margin-left: 3px;
+  }
+}
+@media (max-width: 767px) {
```
dabutvin: can we add the break point to the variables file so we can reference everywhere?

dabutvin: When dealing with the same rule across widths, margin in this case, I suggest we define a base style and then override with either a min-width or a max-width instead of defining max and min around the same breakpoint.

for example

```css
.section-button .text-right * {
  margin: 3px 0 0 0;
}

@media (min-width: $sm) {
  .section-button .text-right * {
    margin: 0 0 0 3px;
  }
}
```

src/styles/_App.scss
```diff
@@ -244,3 +244,18 @@
 hr {
   margin: 75px 0;
 }
+
+.section-header div:nth-child(2) {
+  align-self: flex-end;
+}
+
+@media (min-width: 768px) {
+  .section-button .text-right * {
```
dabutvin: `*` selector seems like overkill for spacing out some buttons here.

```css
.section-button .btn {
  margin-left: 3px;
}
```

src/styles/_App.scss
gianpaj: sure. I'll do that

gianpaj:
Hi @dabutvin , I've made the changes as you suggested

dabutvin:
👍

PR #525, opened 2019-02-28
Improved style of FileList

storrisi:
* added caret icon instead of plus - minus
* Upgraded version of ant-design
* remove line between rows
* increased indent of tree column


dabutvin:
👍

PR #530, opened 2019-02-28
improve perf of browseDefinitionsAction (fixes #528)

### LINKED ISSUE #528: Profile and improve performance of browse page
The network time can be less than 200ms, but the UI can still take a whole second or two to load.
I suspect it is taking a lot of time processing the results on the client.

We should track down and improve the rendering, especially of just landing on /browse

dabutvin:
we were dispatching an update for every definition we got back from the query.
There was also some unused code paths confusing things

### DIFF
--- a/src/actions/definitionActions.js
+++ b/src/actions/definitionActions.js
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: MIT
 
 import { asyncActions } from './'
-import map from 'lodash/map'
 import {
   getDefinitions,
   getDefinition,
@@ -12,7 +11,7 @@ import {
   searchDefinitions
 } from '../api/clearlyDefined'
 import Definition from '../utils/definition'
-import { uiBrowseUpdateList, uiDefinitionsUpdateList } from './ui'
+import { uiDefinitionsUpdateList } from './ui'
 import EntitySpec from '../utils/entitySpec'
 
 export const DEFINITION_LIST = 'DEFINITION_LIST'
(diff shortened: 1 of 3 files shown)


dabutvin:
so fast!

jeffmcaffer:
Anything blocking this smokin' hot performance being unleashed on the public?

src/actions/definitionActions.js
```diff
@@ -99,21 +98,18 @@ export function browseDefinitionsAction(token, query, name) {
...
+        const coordinates = EntitySpec.validateAndCreate(definition.coordinates)
+        if (coordinates) result[coordinates] = definition
+        return result
+      }, {})
+      const allCoordinates = Object.keys(definitionBodies).map(EntitySpec.fromPath)
+      if (query.continuationToken) dispatch(actions.success({ addAll: allCoordinates, data: result.continuationToken }))
+      else dispatch(actions.success({ updateAll: allCoordinates, data: result.continuationToken }))
+
       const definitionActions = asyncActions(DEFINITION_BODIES)
-      definitions.forEach(component => {
```
storrisi: Removing this increase the performance, but adds some problem while operating directly on a specific definition.
If you check the deploy, and try to change something on a definition, you can't actually revert that change using the revert function.

storrisi:
The performance improvement is due to the removal of a portion of dispatching a single action for each definition in the list. But actually that action creates a different part of the store that is used to have a direct reference to the definition in the list.
That is, with this change we are not able to make some operation to specific definitions , like reverting.

Also, i'm in doubt if the removal button has some kind of utility in the browse page?

storrisi:
Update: looks like the changes in this PR has nothing to do with the revert function.
I'm looking at it and i will open an issue together with a separate PR.

PR #536, opened 2019-03-04
normalize file count rendering

jeffmcaffer:
The file counts in the definition views are missing spaces and phrased in the negative

image

Turns out that that code is duplicated in the summary and full details view. This PR factors that out into a shared component and puts it in the positive like this

image


dabutvin:
👍 nice improvement. merging

PR #537, opened 2019-03-05
normalize buttons to OK/Cancel

jeffmcaffer:
Throughout the UI we should be using OK and Cancel (in that order) but we have a mix. This PR fixes the variations.

Note that for some reason in the SourcePicker I am unable to get the styling to have a space between the two buttons. Not sure what our convention is there and how it is happening elsewhere


dabutvin:
👍

PR #538, opened 2019-03-05
switch from component to definition

jeffmcaffer:
The FacetEditor was erroneously referencing a `component` prop when it should have been `definition`


dabutvin:
👍

PR #541, opened 2019-03-05
Issue #499 - Browse: button styling should be updated

### LINKED ISSUE #499: Browse: button styling should be updated
The buttons/dropdowns on the /browse (and other) pages have outlines 

image

whereas it would be much better for them to look like what you see on GitHub itself

image

apply this same change to all lists

storrisi:
Fixes #499


gianpaj:
I've made some progress. Here what's left to update
- [x] Full detail page: Revert Changes button
- [ ] License Expression
- [ ] Browse page
- [ ] Curations page

jeffmcaffer:
can we put these in piece by piece in multiple PRs?

For this specific change, I noticed the Revert button does not show up beside OK/Cancel but does show beside the Files list. Then if you do a change to say the license, an enabled Revert button does show up beside OK/Cancel but the one beside Files is still visible but disabled. What is the expected behavior here.

Note that #499 is all about changing the style of the button in the definition/browse list. Seems you might be addressing other issues here?

jeffmcaffer:
I'd rather not move everything to Ant Design like this. Ant design was brought in to address a small number of cases where we needed some component that fit a need. Fundamentally the app is a bootstrap site. We can of course consider changing that but for that we should have an explicit discussion about the direction of the app as a whole and the look/feel of the various UI frameworks.

Issue #499 is strictly about changing the set of buttons inside the /definitions list (and related lists). Looks like #545 ended up changing the look using just some css changes.

gianpaj:
Sorry. This was my mistake and my lack of requestion clarification via Github or Discord.

I believe this PR be closed as the changes were a lot smaller than I thought

jeffmcaffer:
OK. thanks @gianpaj

PR #545, opened 2019-03-05
update the look of the definition list

dabutvin:
image

image

### DIFF
--- a/src/api/clearlyDefined.js
+++ b/src/api/clearlyDefined.js
@@ -141,17 +141,6 @@ export async function getNotices(token, coordinates, renderer, options) {
   return result
 }
 
-export function getBadgeUrl(score1, score2) {
-  score1 = score1 || 0
-  score2 = score2 || 0
-  const topScore = 100
-  const colors = ['cb2431', 'yellow', '2cbe4e']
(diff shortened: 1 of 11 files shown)


storrisi:
Looks like that some of these fixes have also been addressed in #499 , we should double check the differences.

On which specific issues is related this PR?

jeffmcaffer:
#499 is an issue that mentions some of these changes. So this fixes #499, and perhaps others. Merging this should close #499

dabutvin:
thanks. yeah, some of these ui changes came from just messing around on the fly and weren't fully thought through. I was hoping to address any feedback about the look or feel here in this PR thread.

There are some structure changes here that still need to be reflected in the e2e tests

dabutvin:
actually, looking into this more, the maps are fine the way they are structured. pretty cool.

Any feedback on the way this looks/feels?
[browse](https://deploy-preview-545--clearlydefined.netlify.com/browse)
[fullDetail](https://deploy-preview-545--clearlydefined.netlify.com/definitions/maven/mavencentral/org.flywaydb/flyway-maven-plugin/5.0.7)

PR #550, opened 2019-03-07
Added SpdxPicker in Browse Page filters

### LINKED ISSUE #496: Browse: License list is too restrictive/short
The list of available licenses to filter by is too short. At the very least we should expose all those available from choosealicense.com. That way we are not making a judgement call other than following in their footsteps. That might be too long a list (~36?). Not sure. 

We could instead of a dropdown use the license picker here and let people pick from the full list. would take more real estate but should work reasonably.

Consider a dropdown list like we see in GitHub where the list of choices is long but users can filter by typing at the top. Pretty sure at least one of the component libs we are already using has this capability

storrisi:
Fixes #496 

screen recording 2019-03-07 at 01 52 pm

### DIFF
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -74,10 +74,10 @@ class PageBrowse extends SystemManagedList {
     filterOptions.list = names
     return (
       <Row className="show-grid spacer">
-        <Col md={2} mdOffset={1}>
+        <Col md={1} mdOffset={1}>
           {this.renderFilter(curateFilters, 'Curate', 'curate')}
         </Col>
-        <Col md={8}>
+        <Col md={10}>
           <div className={'horizontalBlock'}>
             {this.renderFilter(providers, 'Provider', 'provider')}
             <span>&nbsp;</span>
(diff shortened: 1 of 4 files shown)


src/components/Navigation/Pages/PageBrowse/PageBrowse.js
```diff
@@ -110,11 +111,7 @@ class PageBrowse extends SystemManagedList {
       <div className="filter-list" align="right">
         <SortList list={sorts} title={'Sort By'} id={'sort'} value={this.state.activeSort} onSort={this.onSort} />
         {this.renderFilter(providers, 'Provider', 'provider')}
-        {this.renderFilter(
-          licenses.filter(license => license.value !== 'absence' && license.value !== 'presence'),
-          'License',
-          'license'
-        )}
+        <SpdxPicker value={''} promptText={'License'} onChange={value => this.onFilter({ type: 'license', value })} />
```
dabutvin: can we get NOASSERTION and blank to show up in this list?

dabutvin:
What do you think about removing the 'No matches found' entry to encourage more freeform text
I think people typing in expressions could be really good use of this textbox.

image

jeffmcaffer:
for now I suggest that we just go with the selections. In the future we should have an expression picker/editor that allows people to either type in an expression or use the more "graphical" design approach we have today. Once we have that we can use it in multiple places including here.

dabutvin:
👍

PR #554, opened 2019-03-07
Browse: Sort by score should use effective and tool 

### LINKED ISSUE #495: Browse: Sort by score should use combined effective score
[changed direction described here]
The /browse page enables people to sort by licensed and described score. We are moving to a model where the pages show just one score, the combined effective score. That will simplify the user model. From the curator point of view that's fine as they should look for things with low score and curate. We'll need to do something else to guide component producers to their components that have low scores.

storrisi:
Fixes #495 

Needs clearlydefined/service#439 to be merged first.

### DIFF
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -90,10 +90,8 @@ class PageBrowse extends SystemManagedList {
     const sorts = [
       { value: 'releaseDate-desc', label: 'Newer' },
       { value: 'releaseDate', label: 'Older' },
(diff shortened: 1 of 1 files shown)


jeffmcaffer:
I added https://github.com/clearlydefined/service/pull/439 to compute the score server side. We also need to add this to the query capabilities.

storrisi:
These are the params that the UI should pass to build the query: https://github.com/clearlydefined/service/pull/439/files#diff-b264c6791ffc7b23ffae5cd68245bbef

storrisi:
I've updated this PR using `effectiveScore` as implemented in clearlydefined/service#439

PR #555, opened 2019-03-07
Browse Page: update name search box to use typeahead

### LINKED ISSUE #548: [Browse] update name search box to use typeahead
see https://github.com/clearlydefined/service/issues/429 for some context

We want to put in exact matches to the search box for names and typeahead should be able to help us with this.

this is rough idea of what I'm thinking

1. Call https://api.clearlydefined.io/definitions/?pattern=jqu like we do from /definitions
2. parse each coordinate response into EntityCoordinates
3. make a distinct list of coordinates.name
4. let the user choose one of the names
5. update the query with ?name=<selection>

This isn't ideal, but may be a good stopgap

storrisi:
Fixes #548 

screen recording 2019-03-07 at 05 29 pm

### DIFF
--- a/src/utils/entitySpec.js
+++ b/src/utils/entitySpec.js
@@ -74,6 +74,10 @@ function normalize(value, provider, property) {
 }
 
 export default class EntitySpec {
+  static isPath(path) {
+    return path.match(/\/*([^/]+)\/([^/]+)\/([^/]+)\/([^/]+)\/?([^/]+)?(\/pr\/.+)?/)
+  }
+
   static fromPath(path) {
     const [, type, provider, namespace, name, revision, prSpec] = path.match(
       /\/*([^/]+)\/([^/]+)\/([^/]+)\/([^/]+)\/?([^/]+)?(\/pr\/.+)?/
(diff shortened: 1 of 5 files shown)


dabutvin:
really nice job!

src/utils/entitySpec.js
```diff
@@ -74,6 +74,10 @@ function normalize(value, provider, property) {
 }
 
 export default class EntitySpec {
+  static isPath(path) {
+    return path.match(/\/*([^/]+)\/([^/]+)\/([^/]+)\/([^/]+)\/?([^/]+)?(\/pr\/.+)?/)
+  }
+
   static fromPath(path) {
```
dabutvin: can we update fromPath to return null if there is no match?

So we don't have to run the same regex twice?

jeffmcaffer:
Love this. Awesome!  Do agree with @dabutvin about the duplicate code. Especially a complex/long regex. Hard to tell if they are actually different or supposed to be the same, or to update. At the very least extract the regex as a constant. Or do what @dabutvin suggested.

Issue #556, opened 2019-03-07
Browse: do not show the Delete and Source buttons in the component list


gianpaj:
As mentioned in the `#implementation` channel in Discord these buttons do not make sense for this page

fossygirl:
I believe this is all implemented and rolled out, so I am closing. Feel free to reopen if you find something different.

PR #558, opened 2019-03-08
Tweak Browse page UI

jeffmcaffer:
Browse page is looking good. Just a couple tweaks

* pull the Curate, Provider and Name inputs to be separate from the list to set a stronger context for the users. Still not fully happy with this. would like to strengthen the Curate messaging so it is clear to users that they can use that to find particular types of things to curate
* removed redundant entry buttons (X and </>) as they cannot be used in the context of the Browse page (system managed list)

image


gianpaj:
This addresses #556

PR #559, opened 2019-03-08
Browse: Make list height relative to height of window

### LINKED ISSUE #498: Browse: Make list height relative to height of window
The current /browse page list looks too short on larger resolution monitors (and possibly too long on lower resolution setups). Rather than trying make one size for all, can we make the list adjust its size based on the height of the window? That would enable people using portrait monitors to get a massive list (which actually could be quite useful).

The current size would be a great minimum height.

This same approach should be used on all the list pages.

storrisi:
Fixes #498

### DIFF
--- a/src/components/ComponentList.js
+++ b/src/components/ComponentList.js
@@ -130,7 +130,7 @@ export default class ComponentList extends React.Component {
     const { loadMoreRows, noRowsRenderer, list, listLength, renderFilterBar } = this.props
     const { sortOrder, contentSeq } = this.state
     return (
-      <div>
+      <div style={{ flex: '1' }}>
         {renderFilterBar()}
         <RowEntityList
           list={list}
--- a/src/components/ComponentList.js
+++ b/src/components/ComponentList.js
@@ -12,7 +12,6 @@ export default class ComponentList extends React.Component {
   static propTypes = {
     list: PropTypes.array,
     listLength: PropTypes.number,
-    listHeight: PropTypes.number,
     loadMoreRows: PropTypes.func,
     onRemove: PropTypes.func,
     onAddComponent: PropTypes.func,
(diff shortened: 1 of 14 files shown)


src/components/ComponentList.js
```diff
@@ -128,16 +127,15 @@ export default class ComponentList extends React.Component {
   }
 
   render() {
-    const { loadMoreRows, listHeight, noRowsRenderer, list, listLength, renderFilterBar } = this.props
+    const { loadMoreRows, noRowsRenderer, list, listLength, renderFilterBar } = this.props
     const { sortOrder, contentSeq } = this.state
     return (
-      <div>
+      <div style={{ flex: '1' }}>
```
storrisi: Can we put this in a css class? Maybe something reusable all over the site where needed

src/components/InfiniteList.js
```diff
@@ -4,11 +4,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { AutoSizer, List, InfiniteLoader } from 'react-virtualized'
-import styles from 'react-virtualized/styles.css'
+
+const noRowsHeight = 200
```
storrisi: Can we put this in the same utils file where we defined other UI elements breakpoint or heights?

gianpaj: The Media queries breakpoints are stored in `_variables.scss`. I can't see which `.js` file you are mentioning in the `/utils` folder. Can you point it to me?

storrisi: Something like you did here https://github.com/clearlydefined/website/pull/484/files/beddda30a246928ae98c5dc959dd93eea3184d2a#diff-be73193a7de1c86c1903f5cf92321d2aR69

gianpaj: I see. done

dabutvin:
👍

Issue #540, opened 2019-03-05
Errors and warnings in console for /definitions


jeffmcaffer:
There are a few errors and warnings in the /definitions page on normal load. We should strive for there to be no errors or warnings (no output at all really). Here is the dump I see. 2 warnings and an error
```
WARNING 1
react-ga.js:204 [react-ga] gaTrackingID is required in initialize()
warn @ react-ga.js:204
_initialize @ react-ga.js:385
initialize @ react-ga.js:429
./src/index.js @ index.js:10
__webpack_require__ @ bootstrap:782
fn @ bootstrap:150
0 @ withTracker.js:38
__webpack_require__ @ bootstrap:782
checkDeferredModules @ bootstrap:45
webpackJsonpCallback @ bootstrap:32
(anonymous) @ main.chunk.js:1


ERROR 1
index.js:1446 Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
    in input (created by AutosizeInput)
    in div (created by AutosizeInput)
    in AutosizeInput (created by HintedInput)
    in div (created by HintedInput)
    in HintedInput (created by TypeaheadInput)
    in div (created by TypeaheadInput)
    in div (created by TypeaheadInput)
    in TypeaheadInput (created by WrappedInput)
    in WrappedInput (created by Typeahead)
    in div (created by Typeahead)
    in Typeahead (created by WrappedTypeahead)
    in WrappedTypeahead (created by WrappedTypeahead)
    in WrappedTypeahead (created by Typeahead)
    in Typeahead (created by OnClickOutside(Typeahead))
    in OnClickOutside(Typeahead) (created by Container)
    in Container (at FilterBar.js:41)
    in FilterBar (at SearchBar.js:11)
    in div (created by Col)
    in Col (at SearchBar.js:10)
    in div (created by Row)
    in Row (at SearchBar.js:9)
    in SearchBar (at PageDefinitions.js:162)
    in div (created by Grid)
    in Grid (at PageDefinitions.js:155)
    in PageDefinitions (created by Connect(PageDefinitions))
    in Connect(PageDefinitions) (at withTracker.js:33)
    in HOC (created by Route)
    in Route (at RehydrationProvider.js:61)
    in Switch (at RehydrationProvider.js:60)
    in main (at App.js:13)
    in div (at App.js:11)
[...]

PR #576, opened 2019-03-11
Added to check to avoid ReactGA run locally

storrisi:
As part of #540 , a check has been introduced to avoid ReactGA to run locally, causing warnings on the console.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,8 +7,10 @@ import './styles/index.css'
 import { unregister } from './registerServiceWorker'
 import { RehydrationProvider } from './components'
 import ReactGA from 'react-ga'
-ReactGA.initialize(process.env['REACT_APP_GA_TRACKINGID'])
-ReactGA.pageview(window.location.pathname + window.location.search)
+if (process.env['REACT_APP_GA_TRACKINGID']) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -7,8 +7,10 @@ import './styles/index.css'
 import { unregister } from './registerServiceWorker'
 import { RehydrationProvider } from './components'
 import ReactGA from 'react-ga'
-ReactGA.initialize(process.env['REACT_APP_GA_TRACKINGID'])
-ReactGA.pageview(window.location.pathname + window.location.search)
+if (process.env['REACT_APP_GA_TRACKINGID']) {
```
gianpaj: wouldn't it be better to check if we're running in `production`?
the `NODE_ENV` should be available:
https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables
>  There is also a special built-in environment variable called `NODE_ENV`. You can read it from `process.env.NODE_ENV`. When you run `npm start`, it is always equal to `'development'`, when you run `npm test` it is always equal to `'test'`, and when you run `npm run build` to make a production bundle, it is always equal to `'production'`.

storrisi: Sure, much better, thanks for the hint

dabutvin:
👍

PR #581, opened 2019-03-12
Removed computedScore() method from the UI

### LINKED ISSUE #578: Remove computeScore() method from the UI
Since now the scores value are returned from the service, we can remove this method and use everywhere the service data.

storrisi:
Fixes #578

### DIFF
--- a/src/components/DefinitionEntry.js
+++ b/src/components/DefinitionEntry.js
@@ -71,7 +71,6 @@ export default class DefinitionEntry extends React.Component {
   renderHeadline(definition, curation) {
     const { namespace, name, revision } = definition.coordinates
     const namespaceText = namespace ? namespace + '/' : ''
-    const scores = Definition.computeScores(definition)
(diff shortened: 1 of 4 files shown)


src/components/DefinitionEntry.js
```diff
@@ -71,7 +71,6 @@ export default class DefinitionEntry extends React.Component {
   renderHeadline(definition, curation) {
     const { namespace, name, revision } = definition.coordinates
     const namespaceText = namespace ? namespace + '/' : ''
-    const scores = Definition.computeScores(definition)
```
jeffmcaffer: perhaps just change this to `const scores = get(definition, 'scores')` then all the other code changes can be reverted and the duplicate code removed. 

The same suggestion applies to `HeaderSection` and `SystemManagedList`

storrisi:
Thanks @jeffmcaffer , nice tip.
I've made the changes.

PR #587, opened 2019-03-13
Switched to Spdx Picker in FilterBar

### LINKED ISSUE #579: Use Spdx Picker as a filter in Workspace page
Since in the Browse page has been implemented the Spdx License Picker as a filter for the list of definitions, we can now use the same component in each page where is needed to filter a list for the same value.

Perhaps we should implement this in the standard FilterBar used for both System Managed List and User Managed List.

storrisi:
Fixes #579


src/components/SystemManagedList.js
```diff
@@ -204,6 +205,7 @@ export default class SystemManagedList extends Component {
   // note that in some scenarios `onFilter` is called with a random second arg
   // and sometimes with an explicit overwrite intent (e.g., true)
   onFilter(value, overwrite = false) {
+    console.log(value)
```
gianpaj: extra console log 🙃

src/components/SystemManagedList.js
```diff
@@ -187,6 +187,7 @@ export default class SystemManagedList extends Component {
       for (let filterType in activeFilters) {
         const value = activeFilters[filterType]
         const fieldValue = this.getValue(definition, filterType)
+        if (!value) return true
         if (value === 'presence') {
           if (!fieldValue) return false
         } else if (value === 'absence') {
```
dabutvin: can we get 'presence' and 'absence' options still?

btw We use this in the 'Share as Notice file' workflow

storrisi: Do you mean to hide that Filter from the list?
I think it could be useful, maybe we should ask for some user feedback on this

dabutvin: I mean with the spdx picker in place we lose the ability to filter by presence of and absence of a license - I think we should retain them, especially with the workflow for share as notice

storrisi: Actually the Spdx Filter we set the same value of the old Dropdown list in the `activeFilters` object.
That should allow the filter `presence` and `absence` still to work.
Did you found a different behavior?

storrisi: I've checked the deploy and i've understand what you mean.
What do you suggest to bring those options back?

storrisi: @dabutvin we need still to solve the mentioned problem here. Do you have any suggestion?

dabutvin: Sorry for the delay.

Let's augment the spdx picker with additional values, the way we do with NOASSERTION/NONE.
It would probably be valuable to separate them by a line.

```
NONE
NOASSERTION
PRESENCE OF
ABSENCE OF
--------------
OBSD
AAL
...
```

jeffmcaffer:
What's the status here? are we good to go?

storrisi:
I've added the custom PRESENCE OF and ABSENCE OF fields to the SpdxPicker used in the FilterBar

dabutvin:
👍 👍 looks good!

PR #588, opened 2019-03-13
Curation component list too short

### LINKED ISSUE #584: curation component list too short
Check out https://clearlydefined.io/curations/1136. Notice that the list of definitions does not grow. In fact, it should be like the others and expand to fill the available space.

On the /definitions page we saw this problem and it was because there was a <div> or some such that did NOT have the `flex-grow` style (or related). Need to ensure that they all have that style or the resize does not work.

storrisi:
Fixes #584


dabutvin:
👍

PR #590, opened 2019-03-14
Show selected value in browse page buttons

### LINKED ISSUE #572: Browse: What selection you've made is not obvious
You may not be able to tell it from the screenshot, but I have currently filtered on Provider="npmjs."

That is not obvious after the filter is applied. The background on the Provider dropdown lit up and the list is only displaying npms. But I could easily see a user forgetting they've applied a filter on this page and not understanding how to un-apply it. 

Screen Shot 2019-03-08 at 8 13 24 PM

storrisi:
Fixes #572

### DIFF
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -18,7 +18,7 @@ import FullDetailPage from '../../../FullDetailView/FullDetailPage'
 import FilterList from '../../Ui/FilterList'
 import SortList from '../../Ui/SortList'
 import ContributePrompt from '../../../ContributePrompt'
-import { curateFilters } from '../../../../utils/utils'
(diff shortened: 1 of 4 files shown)


storrisi:
As discussed in today's weekly meeting, we are going to give to the users an experience pretty similar to GitHub that allows them to know what are the active filters, rather then replace the button of the specific filter with the selected value.

Schermata 2019-03-18 alle 17 55 19

storrisi:
I've updated this one adding a specific "Active Filters" section that show which filters or sorts are applied to the list.
I've factored the implementation as a component so we can maybe think to reuse it somewhere else.

Screen Recording 2019-03-22 at 04 21 PM

storrisi:
I've just added some methods to reset each single filter or even reset all at once

Screen Recording 2019-03-22 at 05 44 PM

dabutvin:
looking good 👍

PR #591, opened 2019-03-14
Browse Page reset search capability

### LINKED ISSUE #573: Browse: Revert changes button not working
Applying a filter of Provider="npmjs" and then clicking the Revert changes button does nothing.

storrisi:
Fixes #573

### DIFF
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -79,11 +79,11 @@ class PageBrowse extends SystemManagedList {
         <Col md={11} mdOffset={1}>
           <div className={'horizontalBlock'}>
             {this.renderFilter(curateFilters, 'Fix something', 'curate', 'success', has(activeFilters, 'curate'), () =>
-              this.onFilter({ type: 'curate', value: null })
+              this.onFilter({ type: 'curate', value: get(activeFilters, 'curate') })
             )}
             <span>&nbsp;</span>
             {this.renderFilter(providers, 'Provider', 'provider', null, has(activeFilters, 'provider'), () =>
-              this.onFilter({ type: 'provider', value: null })
+              this.onFilter({ type: 'provider', value: get(activeFilters, 'provider') })
             )}
             <span>&nbsp;</span>
             <FilterBar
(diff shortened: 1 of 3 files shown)


storrisi:
@fossygirl here you can see how the implementation works:

Screen Recording 2019-03-14 at 07 12 PM

fossygirl:
@storrisi I'm not wild about the x's expanding the selection bar in that way. @dabutvin @jeffmcaffer any thoughts here?

jeffmcaffer:
Agreed. Perhaps we should add "filter badges" along the top of the list (under Browse and on that top row of the list at the left) that have an 'x'. 

At the same time I'm not sure what we are doing here. There is a check mark in the list. Click it again to turn off that selection. The original issue talked about the Revert button but that button is intended to revert changes the user has done to the definitions. It would be odd for it to also affect the filtering. 

As usual, we should ask WWGHD (What would GitHub do) and basically follow their user workflow designs.

storrisi:
>Agreed. Perhaps we should add "filter badges" along the top of the list (under Browse and on that top row of the list at the left) that have an 'x'.

Thanks @jeffmcaffer , this is a good idea and we can implement it if you think it could add some benefits to the user experience

> At the same time I'm not sure what we are doing here. There is a check mark in the list. Click it again to turn off that selection. The original issue talked about the Revert button but that button is intended to revert changes the user has done to the definitions. It would be odd for it to also affect the filtering.

Starting from the issue, the problem that was really addressed is: "As a user, i've done a search, i want to reset the params instead of refresh the page and start again".

Given that, we proposed 2 ways to reach the required behavior:
- expose better the reset capability for each single value (and your comment about showing all the selected values and reset all from there would help for sure)
- add a "clear search" button that resets the entire search.

I've to say that for both the behaviors we have serious problem with Typeahead.

As i mentioned [here](https://github.com/clearlydefined/website/issues/540#issuecomment-473329592) i haven't found the way to clear the selected value for that component with an action called by a button outside of the component itself.
Also, upgrading it will cause introduce some bugs on other points where the same component is used.
I'm taking a look to [https://ant.design/components/auto-complete](https://ant.design/components/auto-complete) but i still have to check if it works with async data, and also if it meets all of our requirements.

storrisi:
As discussed in today's weekly meeting, we are going to give to the users an experience pretty similar to GitHub that allows them to reset all the applied filters.

Schermata 2019-03-18 alle 17 55 19

storrisi:
I'm going to close this one since i've addressed the implementation in #590

PR #596, opened 2019-03-15
Use ListDataRenderer to display a long list of tools

### LINKED ISSUE #547: Tools list cuts off
When we have a bunch of tools run on a component the list gets cut off. We should either wrap the text or enable hover for the full details of the text.

screen shot 2019-03-06 at 11 24 50 am 2

storrisi:
Fixes #547

### DIFF
--- a/src/components/Navigation/Sections/DescribedSection.js
+++ b/src/components/Navigation/Sections/DescribedSection.js
@@ -6,6 +6,7 @@ import get from 'lodash/get'
 import Contribution from '../../../utils/contribution'
 import TwoColumnsSection from '../Sections/TwoColumnsSection'
 import { SourcePicker } from '../..'
+import ListDataRenderer from '../Ui/ListDataRenderer'
 
 class DescribedSection extends Component {
   static propTypes = {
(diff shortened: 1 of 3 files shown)


src/components/Navigation/Sections/DescribedSection.js
```diff
@@ -31,7 +32,10 @@ class DescribedSection extends Component {
       applyCurationSuggestion
     } = this.props
     const definition = Contribution.foldFacets(rawDefinition, activeFacets)
-    const toolList = get(definition.described, 'tools', []).map(tool =>
+
+    // trim tool text if it starts with curation, as what follows is a long hash
+    // e.g. curation/1032ce558db47eee59b75d84132f5fbe6e19efbb
+    definition.described.tools = get(definition.described, 'tools', []).map(tool =>
       tool.startsWith('curation') ? tool.slice(0, 16) : tool
     )
```
dabutvin: I'm a little concerned about mutating the `definition` object here and the ripple effects this might have through the app.

Can the ListDataRenderer take a value instead?

storrisi: I agree, mutating is not the best option here.

I suggest to change this statement in ListDataRenderer:

`const values = get(licensed, item, [])` and accept a property that is the value itself.

gianpaj: I've removed the mutation by passing the value directly to the `ListDataRenderer` as suggested (via an optional prop)

dabutvin:
👍

PR #607, opened 2019-03-18
rename asPrLess to fromObject

jeffmcaffer:


src/utils/entitySpec.js
```diff
@@ -103,7 +103,7 @@ export default class EntitySpec {
     return new EntitySpec(o.type, o.provider, o.namespace, o.name)
   }
 
-  static asPrLess(o) {
```
dabutvin: I think `fromCoordinates` should actually be renamed to `fromObject`

PR #608, opened 2019-03-19
update EntitySpec conversion/trimming names

jeffmcaffer:
the name change done in #607 was less than optimal. Here a set of names are updated to be (hopefully) more consistent.


dabutvin:
👍

PR #621, opened 2019-03-21
Show Tools and Attributions tooltip/overlay on hover

storrisi:
With #596 in place, we noticed that the tooltip shows up only when clicking the Tools or Attributions field, which is not a clear behavior.

We've updated the implementation allow the tooltip to show when hovering the fields.


dabutvin:
👍

PR #611, opened 2019-03-19
Typeahead upgrade

storrisi:
As mentioned in #540 , i've upgraded Typeahead to the latest version.
From here, i've updated some required props all over the components.

I've also split the GitHub Selector to work on 2 separate fields, one for the namespace and one for the package.

Fixes #540

PR #623, opened 2019-03-22
License validation to SPDX Picker

### LINKED ISSUE #620: LicensePicker crashes if the typed license is not valid
When working on the Components list, and you edit a License of a Definition typing some non-valid chars, that are not matching any existing License, then if from there you open the LicensePicker, this will fail returning a runtime error.

I think we should allow the user only to select one of the values from the list of Licenses, avoiding him to type any kind of character.

storrisi:
As discussed in #620 , a validation is required to SPDX Picker to avoid the user to add non-license values.

Fixes #620 

This needs #611 to be in place first.


jeffmcaffer:
this is still in draft. Is it ready to go @storrisi ?

storrisi:
This implementation has been built on top of the upgraded version of Typeahead, so we need #611 to be shipped first

storrisi:
I've fixed some conflicts and now this implementation is ready to be reviewed.

src/components/SpdxPicker.js
```diff
@@ -18,6 +19,35 @@ export default class SpdxPicker extends Component {
...
+    const isMenuOpen = instance.state.showMenu
+    const resultsExist = instance.state.initialItem != null
+    const enterPressed = event.key === 'Enter'
+    // if user is in mid-selection, don't hijack Enter key
+    // i.e. only fire onChange on Enter if menu closed or no results
+    if (enterPressed && (!isMenuOpen || !resultsExist)) {
+      const { target } = event
+      const { value } = target
+      value && this.onChange(value, onChange)
+    }
```
dabutvin: can this validation be updated to support license expressions?

If I enter `MIT OR Apache-2.0` and hit enter, i would like it to work if it is valid

storrisi: It was already in place, but the check for the enter key was avoided by the previous implementation. Now it should be fine.

dabutvin: 👍 👍

PR #629, opened 2019-03-25
Apply Facets on File Lists folders

### LINKED ISSUE #478: Give curators better guidance around "glob expression"
Most of the users who are going to use the Full Details page are not developers. They will not understand what a glob expression is or how to use that functionality. We should change the language on the page to something simpler and also provide tooltips with examples of what is allowed syntax there.

### LINKED ISSUE #562: Need validation on facet user entries
I think it's possible right now to enter anything into a facet glob expression, we need to do validation on the users' entries there so they can't enter bogus stuff.

storrisi:
Fixes #478 
Fixes #562


gianpaj:
Here's a video of this in action (sorry a gif would be too large otherwise)
https://gfycat.com/fixedsizzlingfrillneckedlizard

gianpaj:
I've added two e2e tests.
- 1 for the FacetsEditor
- 1 for the FileList

storrisi:
I'm not sure about why the check are failing here.

The error log tells `error react-scripts@2.1.8: The engine "node" is incompatible with this module. Expected version ">=8.10".` but is the same version of react-scripts used everywhere.

@dabutvin do you have any idea?

dabutvin:
did you ever figure out what is happening with the deployment here and the yarn lock file?

dabutvin:
nice work!

dabutvin:
I think we should remove the 'core' badge from the file list. just show it as blank with the `[+]` button to add new facets

storrisi:
@dabutvin i've fixed the behavior, now it applies facets on all nested folders.

I'm not 100% sure about removing the 'core' facet, it should be a default for all elements that are without a specified facet. 

Thoughts?

dabutvin:
I just want to remove it from the UI, not as a concept.
Since you can't add or remove the core facet directly, i'm worried it will add unneeded confusion/clutter

storrisi:
Don't know, maybe we can involve @jeffmcaffer and @fossygirl in the discussion?

storrisi:
@dabutvin as we agreed, i've removed the `core` default facet from the file list.

jeffmcaffer:
This si looking really good. Love the w

Some feedback, 

* The list of facets is small and fixed. The selection should be a drop down rather than forcing the user to a) know the possible names and b) start typing one of the names
* In the current drop down of suggestions, if you type an `s` you get the following (`examples` is cut off)
Screen Shot 2019-04-30 at 10 43 07 PM
* perf is a bit slow on large definitions with no feedback. Looking at https://deploy-preview-629--clearlydefined.netlify.com/definitions/pypi/pypi/-/grpcio-tools/1.20.1 and setting a facet on the top level folder takes several seconds. the user is left wondering if they did the right thing.
* the list of facets on a given dir goes blank while waiting to process the addition of the second. For exampl,e on the grpcio-tools definition, add `data` to a folder. Then add `dev` to the same folder. While `dev` is being added, the `data` entry disappears and it looks like the folder has no facets.
* The facet glob entry is cut off. Adding the `data` facet to the above mentioned definition gives the following (note that that glob path is actually three segments long.
Screen Shot 2019-04-30 at 10 47 35 PM

storrisi:
@jeffmcaffer thanks a lot for your feedback.

Just a quick answer to:

> perf is a bit slow on large definitions with no feedback. Looking at https://deploy-preview-629--clearlydefined.netlify.com/definitions/pypi/pypi/-/grpcio-tools/1.20.1 and setting a facet on the top level folder takes several seconds. the user is left wondering if they did the right thing.

Once the facet is applied, the preview API is called to return any significative change to the whole definition.
I've tested with the mentioned definition, and it took 3.80s to have a response.

Maybe we can think about a loader, this should allow the user that something is going on, and it will solve also the blank facet problem.

What do you think?

storrisi:
> The facet glob entry is cut off. Adding the data facet to the above mentioned definition gives the following (note that that glob path is actually three segments long.
Screen Shot 2019-04-30 at 10 47 35 PM

This could happen with any very long folder name.
BTW the Ant rendering has also a Tooltip for long texts, which is pretty reasonable in my opinion.

Instead, as in this example, i would remove the "Add a valid glob" and leave more space to the globs:
Schermata 2019-05-02 alle 16 49 52

Thoughts?

dabutvin:
👍 👍 👍

PR #630, opened 2019-03-26
Fix and improve e2e tests

storrisi:
It will be probably easier to review the changes by commits instead of the entire changes

The main change that fixes the test is in ` e2e/tests/definitions.js`:
https://github.com/clearlydefined/website/commit/43dfc5a67e500ad2aab92c76efaad317f9207dde

The other improvements are code style changes (destructuring) and using less code to execute the same tests


storrisi:
We should probably iterate again on this one once the new version of Typeahead will be available.
Some of the DOM elements rendered by Typeahead are changed.

storrisi:
@dabutvin this PR will improve the e2e process, we have several problems during the CI and the fixes implemented here should help

PR #632, opened 2019-03-26
Retry loading a definition entry when it has no data

### LINKED ISSUE #624: Retry loading a definition entry when it has no data
You can add components into a definition list that are not yet harvested. This can happen in a few ways. For example, from dragging and dropping a `package-lock.json` file or a URL to the workspace, or from adding more versions of an existing component.

When this happens, we notice it is missing from our dataset and go off and harvest the component - often times very quickly.

IT is hard to know when the data is available and can require a lot fo manual interaction. Instead of removing and adding the entry from the list to get the new data, the website should check back to the service periodically to see if the definition is available now without any interaction from the user. 

This would allow the user to work through the reset of the components in the list while this happens in the background.

storrisi:
I've added a method launched for the first time when the list definitions has been loaded entirely.
From that, it starts checking for any definition that doesn't contain the `described.tools` property.
If any of that exists, it calls again the getDefinitions method.
Every 20 seconds the check is launched again.
We can of course customize the polling time.

Fixes #624


storrisi:
As discussed in our weekly meeting, i've added here an implementation to show notifications when the harvest process has been started / ended , that fixes #687

storrisi:
@geneh could you please take a look at this and see if everything is fine? Thanks!

src/actions/definitionActions.js
```diff
@@ -29,17 +31,37 @@ export function getDefinitionAction(token, entity, name) {
...
 
+export function checkForMissingDefinition(token, isFirstAttempt = false) {
+  return (dispatch, getState) => {
+    const missingDefinitions = map(
+      get(getState(), 'definition.bodies.entries'),
+      (item, key) => !get(item, 'described.tools') && key
+    ).filter(x => x)
+    if (missingDefinitions.length > 0) {
+      if (isFirstAttempt)
+        uiInfo(dispatch, "Some definitions are not already harvested, we'll notice you once the operation will finish")
```
geneh: Let's change it from
`"Some definitions are not already harvested, we'll notice you once the operation will finish"`
to
`'Some definitions are not already harvested, we'll notice you once the operation finishes.'`

PR #633, opened 2019-03-27
Added sort by lowest Score after a "Fix something" selection

### LINKED ISSUE #593: Add sort after Browse page "Fix something" selection
Once a user has made a "Fix something" selection from the dropdown we should then sort to display lowest-scored components first.

storrisi:
Fixes #593

### DIFF
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -153,7 +153,6 @@ class PageBrowse extends SystemManagedList {
   }
 
   renderFilterBar() {
-    console.log(this.state)
     const sorts = [
       { value: 'releaseDate-desc', label: 'Newer' },
       { value: 'releaseDate', label: 'Older' },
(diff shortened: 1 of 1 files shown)


src/components/Navigation/Pages/PageBrowse/PageBrowse.js
```diff
@@ -126,13 +126,22 @@ class PageBrowse extends SystemManagedList {
...
+    let activeSort = null
     if (overwrite !== true) {
       const filterValue = get(activeFilters, filter.type)
       if (filterValue && activeFilters[filter.type] === filter.value) delete activeFilters[filter.type]
       else activeFilters[filter.type] = filter.value
     }
-    this.setState({ ...this.state, activeFilters: Object.keys(activeFilters).length > 0 ? activeFilters : null }, () =>
-      this.updateData()
+    if (filter.type === 'curate') {
+      activeSort = 'effectiveScore'
```
dabutvin: Could we get the checkmark to display in the sort dropdown when this happens?

storrisi:
Thanks @dabutvin , i've now used the property 'score' which is bound to the Dropdown menu

PR #640, opened 2019-03-29
Score popover text color updates

### LINKED ISSUE #625: Score popover text color updates
before:

image

after:
image

Note: we should still have red badges for low overall scores like we do today, but when you get the popup the text is labeled yellow and green

storrisi:
Fixes #625


dabutvin:
👍

PR #643, opened 2019-04-02
Switch version popup should use montserrat font

### LINKED ISSUE #626: Switch version popup should use montserrat font
before: 
image

after:
image

storrisi:
Fixes #626


dabutvin:
👍

PR #646, opened 2019-04-02
Removed tooltip from Revert button in Full Details view

### LINKED ISSUE #627: tooltip overflows on revert button
When you have edits and open a definition full details, the tooltip is not readable on the revert button

image

It may be easier to just remove the tooltip altogether in this one instance

storrisi:
Fixes #627


dabutvin:
👍

PR #650, opened 2019-04-08
Homepage UI Updates 

### LINKED ISSUE #641: Update UI on homepage
See some of the updates in this picture and I'll try to break them down here

image

- Move 'Active filters' to the Bar above the definition list
- make the filter bar #0000000a
- update the hover on the definition entry to match this color
- the red X to the left of Filter/Sort will clear all filters
- each individual filter can be cleared with it's X
- Make sure sort by dropdown is transparent
- Remove the far-right 'X' from the browse page (Not pictured)
- Veritcally center the footer components (icons, Are you ClearlyDefined?, and terms links)

storrisi:
Fixes #641

### DIFF
--- a/src/components/ComponentList.js
+++ b/src/components/ComponentList.js
@@ -86,7 +86,8 @@ class ComponentList extends React.Component {
       onRemove,
       onRevert,
       showVersionSelectorPopup,
-      hideVersionSelector
+      hideVersionSelector,
+      hideRemoveButton
     } = this.props
     const component = list[index]
     if (!component) return
@@ -118,6 +119,7 @@ class ComponentList extends React.Component {
               getDefinition={this.getDefinition}
               showVersionSelectorPopup={showVersionSelectorPopup}
               hideVersionSelector={hideVersionSelector}
+              hideRemoveButton={hideRemoveButton}
             />
           )}
           onRevert={param => this.revertComponent(component, param)}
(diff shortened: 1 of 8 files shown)


dabutvin:
this is awesome!!!!!

dabutvin:
few notes, nothing major, we can get get this rolled out!

Getting into license:undefined state
image


Can we always show the filter/sort heading? Even when there is nothing selected

storrisi:
Thanks @dabutvin , i've applied the changes.

storrisi:
Thanks @jeffmcaffer :

> Something in this change messes with the spacing of the "Fix something" button. There is intentionally a gap between that button and the Provider and filter. there is also an intentional 1 column width gap at the right of the filter text box. Please restore the spacing and take care in future PRs to retain the spacing.

Thanks for the catch, actually i haven't changed that part in this issue, but i will fix it.

> Is there a fast follow on issue to do these same changes to the Workspace?
> How does this play with the multi-select work (thinking about the counts, the multi-select check box, ...)

I've asked the same question while working on this, see https://github.com/clearlydefined/website/issues/641#issuecomment-480857200

Maybe we could have a second bar in the workspace page, below this filter bar, in order to put there all the multiselect elements?

storrisi:
I've restored the spacing of the buttons in the search section, and also converted "Filters / Sort" to a fixed label.

jeffmcaffer:
Thanks @simo.  Let's drive the workspace up in that other issue. Fwiw, I would not look to add a separate bar to the list. Rather I'd look to use that same space in a similar way.  The lists should look and act VERY similar.

dabutvin:
this is awesome!

PR #659, opened 2019-04-11
Add links for definition files

### LINKED ISSUE #655: Add links for definition files
https://clearlydefined.io/definitions/npm/npmjs/-/cycle/1.0.3

In that definition the `files` entry for `package.json` has a `token` property which is the hash for getting the attached file. The file list entry for `package.json` should be a link that results in the identified attachment is downloaded and showed to the user. Easiest thing for now would be to do that in a simple `text/plain` view in a new browser window.

We have some related function in place for github components but that points to github.

storrisi:
Fixes #655


src/components/FileList/FileList.js
```diff
@@ -124,6 +124,12 @@ export default class FileList extends Component {
     if (!row || !definition) return null
     const { provider, namespace, name, revision } = definition.coordinates
     const path = get(row, 'path')
+    if (row.token)
+      return (
+        <a href={`https://api.clearlydefined.io/attachments/${row.token}`} target="_blank" rel="noopener noreferrer">
+          {row.name}
+        </a>
+      )
     if (provider !== 'github' || !path || row.children) return <span>{row.name}</span>
```
dabutvin: I think we should probably check GitHub first and only use the token if it is not hosted there.

src/components/FileList/FileList.js
```diff
@@ -124,6 +124,12 @@ export default class FileList extends Component {
     if (!row || !definition) return null
     const { provider, namespace, name, revision } = definition.coordinates
     const path = get(row, 'path')
+    if (row.token)
+      return (
+        <a href={`https://api.clearlydefined.io/attachments/${row.token}`} target="_blank" rel="noopener noreferrer">
```
dabutvin: Is there a way to use the current environment prefix the same way we use the definitions API?

Localhost/dev-api/api

storrisi: Sure, we can definitely use the environment var

dabutvin:
playing around with this feature, it's going to be super helpful, I'm excited!

Instead of linking out and losing the formatting, I think we need to take it one step further and open the content in a modal, preserving the formatting of the payload.

what do you think?

storrisi:
It's an option that we should consider.
Which kind of files we are going to show? A Monaco Editor could be useful to show the details?

dabutvin:
probably doesn't need the monaco editor, just a plain `<pre>` tag with a fixed height and a scrollbar for overflow should do it

storrisi:
Got it. Do we need to implement it within this development or we could address this change in a future PR?

dabutvin:
Im worried rolling this out in the current state is going to cause more harm than good, so I'd like to do one more pass to get the formatting of the content before we integrate

jeffmcaffer:
I was talking to Zack from SWH about using SWH to get the file content using the hash urls. Have not had a chance to look at this code but wanted to mention ASAP

dabutvin:
that's awesome!

I think we should have a series of fallbacks for rendering the links.

 - if we resolve github url use that
 - else if we resolve sfh url use that
 - else if we have a token use that
 - else render plain text

jeffmcaffer:
Sounds good. Do you know the format for the SWH commit URLs? @iamwillbar likely does. Also @zacchiro suggested that there is way to augment the URL with file name info so the returned content has the right file name. That would help people doing the download.

iamwillbar:
Here is some prototype code I wrote to get a list of file hashes given a commit hash:
```javascript
function process_revision(callback) {
  const url = `https://archive.softwareheritage.org/api/1/revision/${encodeURIComponent(revision)}/`;
  get(url, (err, body) => {
    if (err) {
      throw err;
    }
    const directory_url = body.directory_url;
    process_directory(directory_url, callback);
  });
}

function process_directory(url, callback) {
  get(`https://archive.softwareheritage.org/${url}`, (err, body) => {
    if (err) {
      return callback(err);
    }
    console.error(hashes.push(...body.filter(item => item.type === 'file').map(item => item.checksums.sha1)));
    async.forEachSeries(body.filter(item => item.type === 'dir'), (item, forEachCallback) => process_directory(item.target_url, forEachCallback), callback);
  });
}

function get(url, callback) {
  request.get(url, {json:true}, (err, response, body) => {
    if (err) {
      return callback(err);
    }
    if (response.statusCode === 429) {
      const waitUntil = parseInt(response.headers['x-ratelimit-reset'], 10);
      console.error(`Waiting until ${new Date(waitUntil*1000).toLocaleString()}`);
      return setTimeout(() => get(url, callback), Math.max(Math.ceil(waitUntil - Date.now()/1000)+1 , 0)*1000);
    }
    if (response.statusCode >= 400) {
      return callback(new Error(`${response.statusCode}: ${response.statusMessage}`));
    }
    return callback(null, body);
  });
}
```

This is the API that would then give the content of a file:
https://archive.softwareheritage.org/api/1/content/raw/

zacchiro:
On Sat, Apr 13, 2019 at 11:40:07AM -0700, William Bartholomew wrote:
> Here is some prototype code I wrote to get a list of file hashes given a commit hash:
>   const url = `https://archive.softwareheritage.org/api/1/revision/${encodeURIComponent(revision)}/`;
[...]
>   get(`https://archive.softwareheritage.org/${url}`, (err, body) => {
[..]
> This is the API that would then give the content of a file:
> https://archive.softwareheritage.org/api/1/content/raw/

The above looks good to me already.

When calling /content/raw/, if you have the filename, you can add the
?filename=... query parameter to make sure that, when saving the file,
the browser will propose the correct filename to the user.  See the doc
of the API endpoint for more details:
https://archive.softwareheritage.org/api/1/content/raw/

Hope this helps,
Cheers

storrisi:
Thanks everybody for the informations, i have only just few questions about the code that @iamwillbar shared.

1. 
> const url = `https://archive.softwareheritage.org/api/1/revision/${encodeURIComponent(revision)}/`;

Probably i'm missing something since i don't know yet how SFW works, what is the source of `revision` param?

2. If i've understand correctly the code, this would return a list of contents retrieved from SFW given a revision commit. I'm wondering if this should an implementation that should be done on the service before return the list of files to the client? I'm thinking on something similar to what we've done with download URLs. Thoughts?

dabutvin:
the url pattern to use for SWH is 

```
https://archive.softwareheritage.org/api/1/content/<token>/raw/?filename=<filename>
```

dabutvin:
Here's an updated proposal, thinking this through more.

For each file we render in the file list:

- if we resolve github url for it, use that
- else link to `https://clearlydefined.io/files/<token>`

This will be a new route + page on the website that renders file content.

From here, use the token from the url to get the content

- if we find it in SWH, use that
- else if we find it in the ClearlyDefined attachments API, use that
- else 404

to call SWH use `https://archive.softwareheritage.org/api/1/content/<token>/raw/`
to call attachments api use `https://api.clearlydefined.io/attachments/<token>`

render the file content in the website chrome, preserving existing formatting.
this text should not be editable

We should put a note at the bottom of the page if it was sourced from SWH
"Content fetched by Software Heritage"

Here's a sample mockup of the look/feel

image

NOTE: we should make sure we follow up on legal obligations around hosting these files this way.

@iamwillbar fyi

storrisi:
@dabutvin it makes sense to me. Do we all agree on this implementation? I'm happy to put in place soon. @jeffmcaffer what do you think?

On a side note, do we have already have some real tokens to test on SWH that return contents?

storrisi:
> else 404

Is that really possible?
If the service returns a token, is there really a chance that both ClearlyDefined or SWH doesn't have the content?

dabutvin:
@storrisi instead of using the token property, we can use `files[n].hashes.sha256`

Every file in the list should be clickable, this is how we will get some 404s.
The idea is that software heritage will have almost everything

storrisi:
I'm doing some tests and we can have different type of contents, of course it depends on the file type.
We can have json, plain texts, markdown, js and so on.

What should be a correct rendering for all of the cases?

dabutvin:
I was hoping we could get away with using a `<pre>` tag

jeffmcaffer:
The approach that @dabutvin outlined makes sense to me. Some comments/answers/questions:

* can/should the service endpoint be /attachments rather than /files? Feels odd to have GET /files/... and no PUT.   Feels like it should be GET /attachments/...?<query arg if needed>
* @storrisi we want to show the content raw. The intent here is for curators to see the content and analyze for licensing info. 
* the 404 case will happen not infrequently for files in packages. SWH has predominately "source control" files. So for example, Maven jars and their contents are unlikely to have hashes in SWH. That's ok. The web ui should handle that and present the user with a gentle but clear message that the file referenced is not available individulally but they can download the full package (we have the download button) to get it all.

storrisi:
Thank @dabutvin and @jeffmcaffer 
You can now check how the behavior works in the deployed version of this PR.
I have still to make some tweaks and checks, but the main functionality is in place.

src/utils/attachments.js
```diff
@@ -0,0 +1,45 @@
...
+    this.revision = revision
+    this.path = path
+    this.row = row
+  }
+
+  getFileAttachmentUrl() {
+    if (this.row.children) return
+    if (this.provider === 'github' && this.path) return this.getGitUrl()
+    else if (this.row.token) return this.getFileRoute(this.row.token)
+    else if (this.row.hashes) return this.getFileRoute(this.row.hashes.sha256)
```
dabutvin: 👍  this is perfect

src/utils/attachments.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  static async fetchAttachmentWithToken(token) {
+    const clearlyDefinedContent = await fetch(this.getClearlyDefinedUrl(token))
+    if (clearlyDefinedContent.status === 200) return clearlyDefinedContent.text()
+    const swhContent = await fetch(this.getSwhUrl(token))
+    if (swhContent.status === 200) return swhContent.text()
```
dabutvin: we need to track whether we got the content from swh or clearlydefined so we can attribute back to them correctly

src/components/Navigation/Pages/PageFiles/PageFiles.js
```diff
@@ -0,0 +1,29 @@
...
+import React, { Component } from 'react'
+import { Grid, Row, Col } from 'react-bootstrap'
+import Attachments from '../../../../utils/attachments'
+
+export default class PageFiles extends Component {
+  state = { content: null }
+
+  componentDidMount() {
+    const token = this.props.location.pathname.slice(this.props.match.url.length + 1)
+    Attachments.fetchAttachmentWithToken(token).then(content => this.setState({ content }))
```
dabutvin: see @jeffmcaffer comment about handling 404 with a message

storrisi: Thanks, see my latest commit

storrisi:
I've added both a message for handling 404 files, and a link showed when a content is fetched from SWH

src/components/Navigation/Pages/PageFiles/PageFiles.js
```diff
@@ -18,9 +18,25 @@ export default class PageFiles extends Component {
...
+                      Content Fetched from{' '}
+                      <a href="https://www.softwareheritage.org/" target="_blank" rel="noopener noreferrer">
+                        Software Heritage
+                      </a>
+                    </span>
+                  )}
+                </>
+              )
+            ) : (
+              <span>The file referenced is not available individulally but they can download the full package</span>
```
dabutvin: @fossygirl @jeffmcaffer we could use a good word-smithing here.

What to tell the user that wants to look at a file that neither swh nor clearlydefined is hosting right now

src/components/Navigation/Pages/PageFiles/PageFiles.js
```diff
@@ -0,0 +1,45 @@
...
+      <Grid className="main-container">
+        <Row>
+          <Col>
+            {content !== 404 ? (
+              content && (
+                <>
+                  <div className="file-content">
+                    <pre>{content.text}</pre>
+                  </div>
+                  {content.source === 'SWH' && (
```
jeffmcaffer: suggest we also have a "Content fetched from ClearlyDefined cache"

src/components/Navigation/Pages/PageFiles/PageFiles.js
```diff
@@ -0,0 +1,45 @@
...
+                    <span>
+                      Content Fetched from{' '}
```
jeffmcaffer: nit lower case "fetched"

src/components/Navigation/Pages/PageFiles/PageFiles.js
jeffmcaffer: Suggest

```
This file is not available individually but you can get it by downloading the full package.
```

For bonus points, "downloading the full package" could be a link that does the same as pushing the Download component button.

src/components/Navigation/Pages/PageFiles/PageFiles.js
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
+// SPDX-License-Identifier: MIT
+
+import React, { Component } from 'react'
+import { Grid, Row, Col } from 'react-bootstrap'
+import Attachments from '../../../../utils/attachments'
+
+export default class PageFiles extends Component {
```
jeffmcaffer: Metaquestion. Should this be `PageFile` (singular). The page will show the content of just one file right?

That should be reflected in the code and the file/directory naming.

src/utils/routingConstants.js
```diff
@@ -10,4 +10,5 @@ export const ROUTE_CURATIONS = '/curations'
 export const ROUTE_DISCORD = '/discord'
 export const ROUTE_STATS = '/stats'
 export const ROUTE_STATUS = '/status'
+export const ROUTE_FILES = '/files'
```
jeffmcaffer: I made a comment somewhere else about the name of the service route. Seems like we don't need a whole new route endpoint, rather just use the /attachments route. Effectively getting attachments with a fallback to use SWH. Just trying to keep the concept count down.

storrisi:
@jeffmcaffer @dabutvin i've update the code with the mentioned suggestions.

If you want i can the download button in the /file page, otherwise we are ready to go

> can/should the service endpoint be /attachments rather than /files? Feels odd to have GET /files/... and no PUT. Feels like it should be GET /attachments/...?

Probably i'm missing something in your sentence, but actually we are already using GET /attachments to retrieve the content

src/utils/attachments.js
```diff
@@ -0,0 +1,46 @@
+import { ROUTE_FILES } from './routingConstants'
+
+// Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
```
jeffmcaffer: Nit but this header should go at the top of the file

src/utils/routingConstants.js
jeffmcaffer: @dabutvin thoughts on this?

jeffmcaffer: Actually, I was confused. This should be singular `/file` as well. The constant name should change too.

storrisi:
CI is hanging after checking successfully e2e tests. Don't know why.

src/components/Navigation/Pages/PageFile/PageFile.js
```diff
@@ -0,0 +1,51 @@
...
+                  <span>
+                    Content fetched from{' '}
+                    {content.source === 'SWH' && (
+                      <a href="https://www.softwareheritage.org/" target="_blank" rel="noopener noreferrer">
+                        Software Heritage
+                      </a>
+                    )}
+                    {content.source === 'CD' && (
+                      <a href="https://clearlydefined.io" target="_blank" rel="noopener noreferrer">
+                        Clearly Defined
```
dabutvin: This should be 1 word

`ClearlyDefined`

dabutvin:
👍 This is good to go!

PR #660, opened 2019-04-15
add stats page to navigation

dabutvin:
show the numbers to the general public

image


dabutvin:
we should probably do #661 first

jeffmcaffer:
Lets look at some font size adjustments before shipping. tehre is a really big differential.

PR #663, opened 2019-04-16
Date picker format consistency

### LINKED ISSUE #638: Date picker switches date format on click
When you are viewing the date picker the date format is listed as YYY-MM-DD and when you click on it it switches to MM/DD/YYY. Let's please do one or the other but not both.

storrisi:
Fixes #638 

Screen Recording 2019-04-16 at 03 45 PM


dabutvin:
👍

PR #664, opened 2019-04-16
Replace "Copy URL" with "Open in a new tab"

### LINKED ISSUE #598: Replace 'Copy URL' with 'Open in new tab'
image

The argument is that the primary reason you are copying the url is to get it open in a new tab, as well, when you open a new tab you have access to the url easily in the address bar if that's what you are after.

image

This shows switching the icon to `fa-external-link-alt`

Clicking it would open a new tab instead of copying the url to the clipboard

storrisi:
Fixes #598


dabutvin:
This is great!

Do you think we can prevent the click propagation so it does not expand the definition entry when clicking the button?

storrisi:
@dabutvin for sure, thanks for the catch

dabutvin:
👍

PR #666, opened 2019-04-16
Replace search by provider with search by type

### LINKED ISSUE #603: In the browse page consider using type rather than provider as a filter
The provider is a proxy for a place. When I search for packages I would typically rather filter by type (npm, maven, etc) than the place they are from.

storrisi:
Fixes #603


dabutvin:
👍

PR #672, opened 2019-04-23
Typeahead Bugfix

storrisi:
I've found a bug on the new released version of typeahead in the LicensePicker.

This will fix that small problem.


dabutvin:
👍

PR #681, opened 2019-04-25
Separate navigationReducer from uiReducer

storrisi:
As discussed with @dabutvin , we need to separate the `navigation` reducer from the others, for 2 reasons:

1. We don't need to persist the `navigation`, the user wouldn't see any change in the menu since the information would be stored in the localStorage
2. Logically, the `navigation` is not a useful information for the workflow of the user, so we can store it in a separate reducer.

I've also added a transform function to the `ui` persistance, in order to store only the `definitions` portion.
This would allow everyone to see this change, otherwise they will continue to take the navigation data from the storage.

### DIFF
--- a/src/components/RehydrationProvider.js
+++ b/src/components/RehydrationProvider.js
@@ -39,6 +39,17 @@ const transformRemoveFetchErr = createTransform(state => omit(state, ['isFetchin
   whitelist: ['session']
 })
 
+// Store only definitions from ui reducer
+const transformUiDefinitions = createTransform(
+  state => {
+    return { definitions: state.definitions }
+  },
+  state => state,
+  {
+    whitelist: ['ui']
+  }
+)
+
 export default class RehydrationDelayedProvider extends Component {
   constructor(props) {
     super(props)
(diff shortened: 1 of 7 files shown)


src/components/RehydrationProvider.js
```diff
@@ -39,16 +39,31 @@ const transformRemoveFetchErr = createTransform(state => omit(state, ['isFetchin
   whitelist: ['session']
 })
 
+// Store only definitions from ui reducer
+const transformUiDefinitions = createTransform(
```
dabutvin: do we need to do this transform if we have a separate navigationReducer now?

storrisi: Yes. It is required to align data for everyone that has already used the persistance of data.
If `redux-persist` will find the `ui` property in the storage, it will use that on a first attempt instead of the real reducer.
Probably, the stored copy will contain also the navigation property, which will never be cleaned.
With this function we will clean that stale information, and also store only necessary data.

dabutvin: Ahhh that makes sense. Thanks

PR #693, opened 2019-05-02
Manually control overlay trigger for Popovers

### LINKED ISSUE #690: Unable to scroll through expanded list of licenses or attributions under definition
**Issue Type:** Bug
**Issue Description:**

Neither the "Declared" or (C) "Attributions" tool-tips allow you to scroll the contents inside them. 

When you hover over a definition's "Declared" or "Attributions" fields, a tool-tip is shown to expand the list of licenses if there are more than a few. However, once you move your mouse toward the tool-tip itself (to manipulate it's scroll bar), the whole tool-tip disappears.

This is most likely also an issue for other fields that use these types of tool-tip.

See: https://clearlydefined.io/definitions/git/github/simplesamlphp/simplesamlphp/92de42cb0a25ccbf46f78a0e2f3257272afb8ba2 for a good example.

storrisi:
Fixes #690 
Screen Recording 2019-05-02 at 05 34 PM


dabutvin:
I'm concerned about all these popovers accumulating just by moving the mouse around.



I wonder if this would work better if it required a 'click' action to launch the popover
then any other click (like for the next popover) would close the previous one

storrisi:
> I wonder if this would work better if it required a 'click' action to launch the popover

We will need something exposed to the user (an icon?) that allows him to know that he need to click to show the popover.

jeffmcaffer:
Is it possible to fix the root cause? Most Hover helps are hover help, not click help :-)

storrisi:
@jeffmcaffer sure. I agree that fixing that would be much better rather than click on an item.

storrisi:
I've tried different solutions but looks like that this issue is much more complicated than expected.
`react-bootstrap` doesn't provides a method to "keep this one open and close everything else".
On the other hand i would like to avoid to pass up props to parents and the again down to components just only to show or hide a popover.
Just thinking to other solutions..

storrisi:
I've found a solution dispatching CustomEvents() and adding a listener on each rendered Tooltip.

dabutvin:
nice work! 👍

Issue #329, opened 2018-10-15
"Pencil" icon for selecting declared license redirects you to a blank page when clicked


bduranc:
Greetings,

When I dig into the following component's definition: 
https://clearlydefined.io/share/eJwtjTEOwyAMRe/imTJE6sLYa1QdLOJItGAjQ1GjiLuHqKzvva9/wBZiJQV3dANFtD52cPyN0YAX0TUwVirgngfUPRM4SNiIwUBWaWG9pn/kiatiHIYxUcnor9pLshuWcfFL0b7Rf4qw9aI0w9FMeptUqYUShIdZ7N0u0F/9BJOCOjI=

, I have observed that the "Pencil" icon used for editing the declared license field redirects me to a blank page instead of actually allowing me to edit the Declared license field.

![2018-10-15_8-17-32](https://user-images.githubusercontent.com/31491646/46960667-e8b77400-d053-11e8-93c6-95cc3b6f5e27.png)

Tested on Chrome 69.0.3497.100, Firefox 62.0.3 64bit, and Edge 42.17134.1.0 and I observe the issue occurs on all three browsers.

Kind Regards,
Brian

dabutvin:
the fix is now live and verified in dev. Will be shipped to prod soon. Thanks all!

bduranc:
@dabutvin: I am still seeing some cases of #329 despite it being closed.

I was trying to declare some licenses on the file level, but I keep getting that infamous blank screen: 

https://clearlydefined.io/definitions/git/github/javaee/javax.annotation/9b07ccb6abe68f55f5d48ecfa84a300998d7c9a0. 

It basically happens whenever I try to edit a field that already has an SPDX expression in it.

Thanks.

storrisi:
@bduranc the problem here is that the file has a license that cannot be validated by the spdx library that we use, since the license contains a NOASSERTION value: `CDDL-1.1 OR NOASSERTION AND GPL-2.0-only`

So i've made a quick update to avoid this, but i'm wondering if this is really possible?

bduranc:
@storrisi : I'm not sure I really follow. In almost all cases, it seems these are files that already have a valid SPDX expression entered of either NOASSERTION by itself or in combination with other licences.

I simply wished to update the license to a different value. The expression I planned to enter would have also been a valid SPDX. Other fields let me do this without any trouble, even if there is a value already existing in the field.

storrisi:
[...]

PR #695, opened 2019-05-06
Fix license parsing when it contains NOASSERTION

storrisi:
As mentioned in https://github.com/clearlydefined/website/issues/329 , an error occurs when trying to editing a license that contains a NOASSERTION value.

It wont be validated from the SPDX library, so i've added an additional check.


src/components/LicensePicker/utils.js
```diff
@@ -10,7 +10,12 @@ const NOASSERTION = 'NOASSERTION'
 // Shared methods appliable to License Picker
 export default class LicensePickerUtils {
   static parseLicense(license) {
-    return license && !['NONE', 'NOASSERTION'].includes(license) ? parse(license) : { license }
+    return license &&
+      !['NONE', 'NOASSERTION'].includes(license) &&
+      !license.includes('NOASSERTION') &&
+      !license.includes('NONE')
```
dabutvin: Do we need all 3 of these conditions? Can we drop the first one now?

src/components/LicensePicker/utils.js
```diff
@@ -10,7 +10,12 @@ const NOASSERTION = 'NOASSERTION'
...
+      ? parse(license)
```
dabutvin: Is it parse() that does not handle NOASSERTION?

We might want to sync our spdx parsing across service and website

storrisi: Yes, parse() fails when it contains a NOASSERTION value.
Has this been already solved on the service? I'm gonna check then.

storrisi: Are you referring to https://github.com/clearlydefined/service/blob/master/lib/spdx.js#L21 ?

dabutvin: yep. 
I put up a repo at https://github.com/clearlydefined/spdx if we want to experiment with sharing the code from that spot

storrisi: What about the way to share the code? Should we use a simple `npm link` instruction or maybe publish a library on npm?

dabutvin: we can publish to npm. `@clearlydefined/spdx`

src/components/LicensePicker/__tests__/LicensePickerUtils.test.unit.js
```diff
@@ -23,9 +23,9 @@ describe('LicensePickerUtils', () => {
       'MIT-0': true,
       'MIT AND Apache-2.0': true,
       'MIT OR Apache-2.0': true,
-      junk: false,
+      junk: true,
```
dabutvin: should the string `junk` be considered valid?

is it when it is invalid that is crashes the control?

storrisi: This is a change introduced by the usage of the custom Spdx validation lib.
Looks like that if any kind of wrong license is passed, then it will considered as valid since it would be changed in `{noassertion: null}`

dabutvin: i see. makes sense

package.json
```diff
@@ -34,7 +34,7 @@
     "redux": "^3.7.2",
     "redux-persist": "^4.6.0",
     "redux-thunk": "^2.3.0",
-    "spdx-expression-parse": "^3.0.0",
+    "spdx": "github:clearlydefined/spdx",
```
dabutvin: you can update this to `"@clearlydefined/spdx": "0.1.0"`
https://www.npmjs.com/package/@clearlydefined/spdx

storrisi: great! will do.

dabutvin:
👍

dabutvin:
opened https://github.com/clearlydefined/service/issues/517 to track getting this into the service

PR #703, opened 2019-05-08
Fix Typeahead behavior with Git commits

### LINKED ISSUE #701: Suggestions not showing up in source location dialog
**Issue Type:** Bug
**Issue Description:**

The functionality mentioned in #179 appears to be experiencing a major issue since the last major update to it sometime back in March or Apr.

When one clicks the source location field on a definition, this results in a pop-up window and various fields for the repo (at this moment: GitHub org/name, repo, and commit hash).

Normally, all of these fields should present the user with various recommendations as they type:

Example1

However, when it comes to enter in the commit hash, the items of recommendation do render but without any text. So to the user, it looks blank:

Example2

I.e. If one clicks on the area shown in red (a version recommendation for the commit hash entered), it **will** populate with the version. However, the text for that item is **not visible in the list**.

Tested in Firefox 66.0.3 (64) and Chrome 74.0.3729.131 (64).

storrisi:
Fixes #701


dabutvin:
can we remove the `(undefined)` from showing up for commit hash:

image

storrisi:
@dabutvin sure! I'm happy to fix that, could you tell me which data did you insert in order to reproduce it and update tests as well?

storrisi:
Nevermind, i found that it happens any time you type something.

dabutvin:
looks good 👍

PR #712, opened 2019-05-14
Move facets tooltip near section title

### LINKED ISSUE #706: group the tooltip for facets with the title
image

The (i) is aligned with the second column, should be right against the word 'Facets'

storrisi:
Fixes #706


dabutvin:
:+1:

Issue #652, opened 2019-04-09
Improvement to score metric experience


bduranc:
**Issue Type:** Feature Request / Improvement to existing feature

**Issue Description:** The license-score is critical to being able to quickly determine the completeness and accuracy of a CD component. However, it may be difficult for newcomers (and perhaps even some experienced users) to interpret as currently displayed in the UI.

These suggestions on how to improve the readability of the score is two fold, covering context and general UI enhancements. We'd do this for both the "Described" and "Licensed" scores. However, this proposal primarily focuses on just the "Licensed" score.

**Proposal:**

`1. Add a "/scoreHighBounds" suffix to each of the score metrics.`

Why? The score is out of 100, however at first glance no indication of a range bounding is given. This may cause people to ask themselves: Is this just a count of licenses/things found or is it actually a calculated score? What is the difference between "Effective" and "Tools" columns? To those involved intimately with CD, these questions perhaps have definitive answers, but to newcomers it may not be precisely clear.

`2. Add a "layman terms" description link to a simplified version of the score metrics document:`

The document at https://github.com/clearlydefined/license-score/blob/master/ClearlyLicensedMetrics.md explains the score criteria but is a fair bit to read through.

A simplified version of this document could be created in the same docs repository and we put a link to it at the bottom-right corner of the score tool-tip in the UI.

For this to work, we also need to ensure the documents and naming of score metrics are consistent with how they are named in the UI. I.e. ("Discovered" metric in UI currently corresponds to "File-level license and copyright" metric in the docs)

`3. Decrease score tool tip transparency and change "red" to "orange".`

The fact that the current tool tip is somewhat transparent, means it blends in with other background text and becomes difficult to read. The current text color for "red" is also hard on the eyes.

For these reasons, I propose...

a) we make the following changes to the tool tip: change the RGB alpha on the tool tip background color to .90%. up from the current of .75%. This will keep it somewhat opaque, but not enough that still blends in with other background text.

b) we change the "red" ( rgb(203, 36, 49) ) text color for each metric to a dark yellow such as hex #eaa927.

**Mock-up below: (styling values used are noted above)**

<img width="722" alt="2019-04-01_12-37-51" src="https://user-images.githubusercontent.com/31491646/55824836-877bd500-5ab9-11e9-8e69-fc52cd204d91.png">

bduranc:
Two other post-submission thoughts that came up but were not included in the original proposal (or the mock-up), so here they are below:

[...]

PR #713, opened 2019-05-14
Score Rendering improvements

storrisi:
This is related to #652 

This is how the tooltip looks like after the suggested changes:
Schermata 2019-05-14 alle 18 36 54

This implementation still miss one of the suggested changes, about link the titles to a document, that is not yet be written.

### DIFF
--- a/src/components/Navigation/Ui/ScoreRenderer.js
+++ b/src/components/Navigation/Ui/ScoreRenderer.js
@@ -49,7 +49,8 @@ class ScoreRenderer extends Component {
     const color = this.getColor(score[name], maxScores[name] || 100, ['#d6af22', '#2cbe4e'])
     return (
       <p style={{ color }}>
-        {label}: {score[name]}/100 points
+        {label}: {score[name]}
+        {this.renderUnit()}
       </p>
     )
   }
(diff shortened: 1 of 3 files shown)


storrisi:
I've also added a link to the documentation which completes and fixes #652

src/components/Navigation/Ui/ScoreRenderer.js
```diff
@@ -71,28 +72,56 @@ class ScoreRenderer extends Component {
...
+          </h2>
           {this.renderScore(tools)}
         </div>
       </div>
     )
   }
 
+  renderUnit = () => (
+    <span>
+      / 100 <span className="unit">points</span>
```
geneh: Let's drop "points" everywhere and make sure the spacing around the slash is even.

src/components/Navigation/Ui/ScoreRenderer.js
```diff
@@ -71,28 +72,56 @@ class ScoreRenderer extends Component {
...
+    </span>
+  )
+
+  renderScoreInfo = () => (
+    <div className="ScoreInfo">
+      <a
+        href="https://github.com/clearlydefined/license-score/blob/master/ClearlyLicensedMetrics.md#clearlylicensed-scoring-formula"
+        target="_blank"
+      >
+        What does these scores mean?
```
geneh: Let's change it to "Scoring formula"

geneh:
👍

PR #716, opened 2019-05-16
introduce 'OTHER' as a choice for curating license

dabutvin:
This is the idea:
- NOASSERTION is a machine only term
- curations can choose from NONE / OTHER / SPDX

This is meant to short circuit the curation infinite loop.
If you pick up a component that is NOASSERTION, you can mark it as NONE or OTHER and move on

---

### picking a license for curation:

image

### filtering on homepage:

image

### filtering on workspace:

image


dabutvin:
@royaljust fyi

PR #724, opened 2019-05-22
Expand files list by default

### LINKED ISSUE #651: It is hard for curators to find what files need curation
It is very difficult for a curator to find what files need to be curated in a given file system.  For example, filtering by "NOASSERTION" in apache spark requires 6 clicks to see the first three files to curate; and 10 clicks to get to the next file.

For a given component, giving a flat list of files in need of curation would be very beneficial.

image

storrisi:
This resolves #651 and expand the entire file list by default.

BTW, there are some performances issues when playing with a very long list of files.
I will track that problem in another PR.

### DIFF
--- a/src/components/FileList/FileList.js
+++ b/src/components/FileList/FileList.js
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation and others. Licensed under the MIT license.
 // SPDX-License-Identifier: MIT
 
-import React, { Component } from 'react'
+import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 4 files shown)


dabutvin:
Is it possible to tweak this so that it only expands everything for a filtered list?
And then collapse everything when a filter is cleared?

Thoughts on that interaction?

storrisi:
@dabutvin sure we can. I've to make some tests about performances, since any operation done to a long list of files takes a while to be rendered.

storrisi:
I've updated the behavior as expected. We should do more investigations about performances, but we can track it on another PR.

PR #736, opened 2019-06-05
Added suggestions to the License Picker modal

### LINKED ISSUE #705: Move license suggestions into the spdx picker
right now the suggestions are next to the license picker as an 'i'

image

It would be great to add them to the top of the spdx picker when you open it instead.

Similar to how we separate the NONE/NOASSERTION in the filterbar, we can separate and label the suggestions.

```
--------------------------
suggested
-----------
MIT (version 1)
Apache-2.0 (version 2)
all
------------
NONE
NOASSERTION
0BSD
AFL
...   
```

(all and suggested would be lighter font)
suggested and all may need to have different padding to make it clear they are grouped

storrisi:
Screen Recording 2019-06-05 at 06 22 PM

Fixes #705

### DIFF
--- a/src/components/LicensePicker/LicensePicker.js
+++ b/src/components/LicensePicker/LicensePicker.js
@@ -43,10 +43,23 @@ export default class LicensePicker extends Component {
       this.setState({
         ...this.state,
         licenseExpression,
-        rules: LicensePickerUtils.parseLicense(licenseExpression),
         isValid: LicensePickerUtils.isValidExpression(licenseExpression)
       })
     }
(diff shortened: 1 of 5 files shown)


src/components/LicensePicker/LicensePicker.js
```diff
@@ -81,13 +95,18 @@ export default class LicensePicker extends Component {
   }
 
   removeRule = async rule => {
+    console.log(rule)
```
geneh: Let's remove console.log

PR #760, opened 2019-06-18
Increased size of revision string

### LINKED ISSUE #753: 4th version part of NuGet being dropped off in Workspace list
If you add the component AWSSDK.Core 3.3.101.1 to the workspace it is listed as 3.3.101. This could be ambiguous if you added multiple versions that only differed by the fourth part.

storrisi:
Resolves #753

### DIFF
--- a/src/components/DefinitionEntry.js
+++ b/src/components/DefinitionEntry.js
@@ -93,7 +93,7 @@ class DefinitionEntry extends React.Component {
     return (
       <span>
         <DefinitionTitle definition={definition} />
-        <DefinitionRevision definition={definition} />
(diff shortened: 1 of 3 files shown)


storrisi:
@geneh slicing the revision is a change added to avoid very long revision strings in the Definitions list in a mobile view, like the one provided from github (ex. d03f9f82e7dd3511d428056a722c41eb6e1fa868 )

Maybe we can change this behavior and give to the string a fixed width, and then in the detail show the entire string.

What do you think?

storrisi:
I've made a change as explained above. I've also added a Tooltip so if there is a very long revision string, then the user could always check it.

PR #772, opened 2019-06-26
Debian integration

### LINKED ISSUE #767: Add website support for Debian package & source
Add website changes required for supporting Debian package & source.
More details can be found at https://github.com/clearlydefined/crawler/issues/321.

storrisi:
Resolves #767


storrisi:
We're starting to have too much buttons on the harvest page:
Schermata 2019-07-01 alle 16 36 01

Perhaps we could think on using logos instead of names? 

@fossygirl thoughts?

src/api/clearlyDefined.js
```diff
@@ -21,14 +21,16 @@ export const ORIGINS_CRATE = 'origins/crate'
...
 export const ORIGINS = {
   github: { git: ORIGINS_GITHUB },
   npmjs: { npm: ORIGINS_NPM },
   nuget: { nuget: ORIGINS_NUGET },
   cratesio: { crate: ORIGINS_CRATE },
   mavencentral: { maven: ORIGINS_MAVEN, sourcearchive: ORIGINS_MAVEN },
   pypi: { pypi: ORIGINS_PYPI },
-  rubygems: { gem: ORIGINS_RUBYGEMS }
+  rubygems: { gem: ORIGINS_RUBYGEMS },
+  debian: { gem: ORIGINS_DEBIAN }
```
geneh: deb instead of gem?

src/components/FilterBar.js
```diff
@@ -22,7 +23,8 @@ const types = {
   nuget: nuget,
   git: git,
   crate: crate,
-  pod: pod
+  pod: pod,
+  debian: debian
```
geneh: The path is going to be deb/debian/... Not sure if it should be deb here.

src/utils/__tests__/entitySpec.test.unit.js
```diff
@@ -86,4 +86,14 @@ describe('EntitySpec', () => {
     const result = EntitySpec.fromUrl('https://rubygems.org/gems/tzinfo/versions/1.2.5')
     expect(result.toString()).toEqual('gem/rubygems/-/tzinfo/1.2.5')
   })
+
+  it('returns a debian path with a version', async () => {
+    const result = EntitySpec.fromUrl('https://sources.debian.org/api/src/amoeba/1.1-30/')
+    expect(result.toString()).toEqual('deb/debian/-/amoeba/1.1-30')
```
geneh: For Debian packages there is going to be a namespace, for example deb/debian/amd64/amoeba/1.1-30.

storrisi: Do you have some rules that i can use to build a proper parser?
For example, using the URL above, how do you know that the namespace should be amd64 ?

storrisi:
> There is also going to be a new source type debsrc since the sources are there.

@geneh i'm not sure what this could mean for the website, probably it's related for the sources of the files in a definition?

src/components/FilterBar.js
storrisi: This is only used to get the logo, so it wont be involved by the path structure here

src/utils/__tests__/entitySpec.test.unit.js
geneh: The namespace is the architecture. As with other name-spaced packages, the users should be able to provide it. Does it make sense?

geneh:
> > There is also going to be a new source type debsrc since the sources are there.
> 
> @geneh Gene Hazan FTE i'm not sure what this could mean for the website, probably it's related for the sources of the files in a definition?

Yes. It's a kind of similar to Maven's source archive type.

src/utils/__tests__/entitySpec.test.unit.js
storrisi: @geneh not totally. What does should change from a curator point of view? Should the user really know if he's curating a package on a specific architecture?
Probably i'm missing something.
@iamwillbar thoughts?

geneh:
@storrisi Could you please resolve conflicts and check if everything makes sense given https://github.com/clearlydefined/service/pull/612?

PR #793, opened 2019-07-29
Avoid set null value on source location object

### LINKED ISSUE #792: Closing the source location editor with OK adds a bogus entry in the curation
#### Steps to reproduce:
1. Open a definition
1. Click on the source location so that the source location editor opens as a popup (many curators do that so they can use the anchor)
1. Click OK
1. Make changes to the definition
1. Try to submit the changes

#### Expected result:
- Curation created

#### Actual result:
- Error: source location needs to be an object.

storrisi:
Fixes #792


geneh:
@storrisi, the changes in package-lock file are breaking the build, so I had to revert the changes from master. Can you please create another PR just with the changes in src/components/SourcePicker.js?
Are you using Node v10 and NPM v6?

PR #824, opened 2019-09-10
Add matchingFalse param for definitions request

### LINKED ISSUE #823: Add a matchCasing=false parameter when calling the /definitions endpoint
As explained on https://github.com/clearlydefined/service/issues/589#issuecomment-529936105 and because of what we found as #604 , we need the Website to keep using a non-post-processed response.
Please add a query string parameter ```matchCasing=false```, on requests to the /definitions endpoint which end up composing curations. Alternatively, just replace on all usages of that endpoint.

storrisi:
Fixes #823

### DIFF
--- a/src/api/clearlyDefined.js
+++ b/src/api/clearlyDefined.js
@@ -109,7 +109,7 @@ export function getDefinition(token, entity, params = {}) {
   return get(
     url(`${DEFINITIONS}/${entity.toPath()}`, {
       expand: expandPrs ? 'prs' : null,
-      matchCasing: false
+      matchCasing: 'false'
     }),
     token
   )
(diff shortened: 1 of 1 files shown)


geneh:
Should E2E or other tests be updated? Let's deploy both the website and the service changes together and test to make sure nothing gets broken.

storrisi:
@geneh E2E and tests don't use real endpoints, they just mock the response.

geneh:
Okay. I've been playing with Netlify deployment https://deploy-preview-824--clearlydefined.netlify.com and definitions endpoints do not seem to have the new parameter. It was deployed more than three hours ago, so it shouldn't be a caching issue.
@storrisi Can you point me to a page where the new parameter is expected to show up?

storrisi:
@geneh you were right, thanks to your tests I discovered that the matchCasing param should be passed as a string.
Since its value was a boolean false, it was never added to the query string.

geneh:
Great, thanks!

PR #838, opened 2019-10-23
Add deb and debsrc stats

### LINKED ISSUE #836: Debian numbers are off on the stats page
https://clearlydefined.io/stats
Debian total count is 0 which is incorrect.
The search type is currently **debian** which is incorrect. It should be **deb** for Debian packages and **debsrc** for Debian sources.
Please correct and also add stats for Debian sources.

storrisi:
I've modified Debian stats in order to check for `deb` param.
I've also added a card for `debsrc`, but the call to the service fails:
Schermata 2019-10-23 alle 10 05 36

Fixes #836

### DIFF
--- a/src/components/Navigation/Pages/PageStats/PageStats.js
+++ b/src/components/Navigation/Pages/PageStats/PageStats.js
@@ -28,7 +28,8 @@ const types = {
   nuget: nuget,
   git: git,
   crate: crate,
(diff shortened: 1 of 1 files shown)


geneh:
Added https://github.com/clearlydefined/service/pull/647 to address.

PR #847, opened 2019-11-04
Fix npm outdated dependencies

### LINKED ISSUE #842: Upgrade dependencies with security vulnerabilities
Ideally all `npm outdated` dependencies and at the very least all vulnerable components (`npm audit`) should be upgraded.

storrisi:
Fixes #842


geneh:
There are no updates to the dependencies here. Can they be added?

storrisi:
It looks that the reverted changes removed my upgrades. Let me investigate.

geneh:
@storrisi Any update on this?

storrisi:
I'm still getting errors while trying to run `npm audit fix` . Just trying to figure out why

geneh:
Which package is causing it?

storrisi:
The log doesn't explain which is, I'm trying to remove and adding back dependencies to see which is the one that fails

storrisi:
@geneh i think i found the lib that cause the problem. Apparently is `@clearlydefined/spdx`

storrisi:
Ok, after a lot of tries, i can confirm that the problem is on `@clearlydefined/spdx` lib.
As soon this lib is added to the project, the `npm audit` command starts failing.
Maybe it has some deps to be updated as well?

geneh:
Good catch. Yes, it's possible. Luckily, we own this repository. Could you please check?

storrisi:
@geneh sure i will

geneh:
Is the code ready to be merged?

storrisi:
Yes it is!

geneh:
Great, thanks!

PR #852, opened 2019-11-18
Add chart showing tools ran per day

### LINKED ISSUE #774: Show a bar graph on /status page for tools
Would be nice to add a bar graph on clearlydefined.io/stats that shows the number of tools that have been run each day. 

@iamwillbar

storrisi:
Fixes #774

### DIFF
--- a/src/components/Navigation/Pages/PageStatus/PageStatus.js
+++ b/src/components/Navigation/Pages/PageStatus/PageStatus.js
@@ -46,7 +46,8 @@ export default class PageStatus extends Component {
       this.fetchDefinitionAvailability(),
       this.fetchCrawledPerDay(),
       this.fetchRecentlyCrawled(),
-      this.fetchCrawlbreakdown()
+      this.fetchCrawlbreakdown(),
+      this.fetchToolsranperday()
     ])
     this.setState({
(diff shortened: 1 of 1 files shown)


src/components/Navigation/Pages/PageStatus/PageStatus.js
```diff
@@ -46,15 +46,17 @@ export default class PageStatus extends Component {
       this.fetchDefinitionAvailability(),
       this.fetchCrawledPerDay(),
       this.fetchRecentlyCrawled(),
-      this.fetchCrawlbreakdown()
+      this.fetchCrawlbreakdown(),
+      this.fetchToolsranperday()
```
geneh: Let's change it to camel case here and everywhere `this.fetchToolsRanPerDay()`.

geneh: @storrisi Does it make sense? Is there a particular reason for this casing?

storrisi: Sorry for the delay. I just copied the value coming from the backend, your change request definitely makes sense for consistency.

PR #868, opened 2020-02-24
Make sure the right casing is used when a new curation is created

geneh:
Related to https://github.com/clearlydefined/service/issues/705


jeffmendoza:
LGTM

PR #928, opened 2022-01-28
Error fixes

andrewrmn:
This PR fixes the errors and broken links from the new-ui branch.

PR #932, opened 2022-02-04
removes e2e tests

nellshamrell:
Several of these tests broke at some point in the past and we have not been able to fix them.

Additionally, once #928 is deployed (which is our new UI), they will be further broken.

I'm certainly open to re-adding end to end tests like these at some point in the future, but this will unblock deploying the new UI.

Signed-off-by: Nell Shamrell <nells@microsoft.com>


nellshamrell:
I have disabled the e2e pipeline in Azure DevOps as well.

disulliv (approved):
I think as long as we open an issue to re-implement them at some point, this is fine, especially since it's a "blocker" on the new UI.

nellshamrell:
@disulliv good point, opening an issue now.

Issue #955, opened 2022-03-01
Not able to edit file-level license


capfei:
The edit button does nothing when trying to edit the license info for a file. Also, clicking on the license field does nothing.
<img width="823" alt="image" src="https://user-images.githubusercontent.com/5489711/156204040-c59ce117-4cc4-40f8-be75-ef5173ea54e2.png">

capfei:
Can confirm that I am able to do file-level edits (haven't submitted a PR, though) and experiencing the same issue with the radio buttons.

qtomlinson:
Commit `https://github.com/clearlydefined/website/pull/977/commits/5c0a7215567b1506395c8b375edb680e19f62f65` reused LicensePicker for license editing.  It seems that LicensePicker needs some work as well.

qtomlinson:
@capfei @bduranc The improvement needed in LicensePicker is more general and not limited to "file-level license" editing. Editing license in the "Described and Licensed" of the definition page shows the same dialogue and needs the same improvement.  The original use case of "Not able to edit file-level license" is now supported.  The improvement LicensePicker  is better documented in a separate issue, so that all the test cases (file-level and declared license) can be addressed.

![image](https://user-images.githubusercontent.com/92891533/211618599-bf5c3e74-fd14-4f97-bc9a-832f897079a6.png)

bduranc:
Alright, I will open new issue to discuss the usability of this dialog box.

It is worth mentioning that in the old UI, this function did work much better because none of these issues existed.

bduranc:
I confirmed I was able to submit this curation PR: https://github.com/clearlydefined/curated-data/pull/23164

I created issue #1013 addresses the larger problemswith the LicensePicker component itself. But for the original scope of this issue, it looks like the problem is fixed.

@capfei do we want to close this?

PR #980, opened 2022-05-09
Support license edit

qtomlinson:
Task: https://github.com/clearlydefined/website/issues/955

### DIFF
--- a/src/components/FileList/FileList.js
+++ b/src/components/FileList/FileList.js
@@ -237,7 +237,7 @@ export default class FileList extends PureComponent {
     onChange(`described.facets.${facet}`, newGlobs)
   }
 
-  onLicenseChange(record, license) {
+  onLicenseChange = (record, license) => {
     const { onChange, component, previewDefinition } = this.props
     const attributions = Contribution.getValue(component.item, previewDefinition, `files[${record.id}].attributions`)
     onChange(`files[${record.id}]`, license, null, license => ({
@@ -293,7 +293,7 @@ export default class FileList extends PureComponent {
     )
   }
 
-  renderLicenseCell(record) {
+  renderLicenseCell = (value, record) => {
     const { readOnly, component, previewDefinition } = this.props
     const field = `files[${record.id}].license`
     const editor = EnhancedLicensePicker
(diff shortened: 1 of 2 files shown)


qtomlinson:
@disulliv @mpcen ready for review :)

disulliv:
Code looked good to me! @qtomlinson do you have any screenshots from testing it?

qtomlinson:
Before Edit
image
Edit
image
After Edit
image

src/components/FileList/FileList.js
```diff
@@ -235,6 +237,16 @@ export default class FileList extends PureComponent {
     onChange(`described.facets.${facet}`, newGlobs)
   }
 
+  onLicenseChange(record, license) {
```
mpcen, line 240: nit: You can get some perf increases for free by binding this method here rather than inside of render on ln: 310. 

Also, FWIW, most, if not all of the methods of this class are bound in the method definition using arrow functions instead of binding in render.

src/components/FileList/FileList.js
```diff
@@ -281,6 +293,25 @@ export default class FileList extends PureComponent {
     )
   }
 
+  renderLicenseCell(record) {
```
mpcen, line 296: Same binding nit

mpcen (changes requested):

src/components/FileList/FileList.js
qtomlinson: Agreed.  Using arrow function is more readable, e.g. `onLinceseChange = (record, license) =>{}`

Inside renderLicneseCell on line 310, however, parameter `record` still needs to be passed in.  Looks like two options:

 `onChange={license => this.onLicenseChange(record, license)}`

or

 `onChange={this.onLicenseChange.bind(null, record)}`

So binding in the render is still necessary.

src/components/FileList/FileList.js
qtomlinson: changed to use arrow function instead. See https://github.com/clearlydefined/website/pull/980/commits/ea7609b1515cce98401db78882df1a4c788e3872

mpcen (approved):

Issue #957, opened 2022-03-01
Search filter does not filter type


capfei:
Tried searching for npm package called "[cli](https://www.npmjs.com/package/cli)", but search suggestions show everything even though filter is set to "NpmJs". Same result if I change it to different filters. Same behavior in Chrome and Edge.

<img width="698" alt="image" src="https://user-images.githubusercontent.com/5489711/156232207-b312f1d2-48f0-4727-a2e6-6740cea91a35.png">

bduranc:
I took a look at this in dev environment. 

Although the search relevance has certainly improved, I still observed one or more cases where other package types are shown (for example, you select "GitHub" from the dropdown but still get NPM packages in the mix too).

![image](https://user-images.githubusercontent.com/31491646/193114419-78eae779-8a37-4445-9edf-01f7c9f20e93.png)

Furthermore, when I actually select the package from the search result it may not always show in the list below. 

This is certainly still the case with more generic named (NPM) packages like "_cli_" or "_log_", but does not appear to impact more specific searches such as "_cli_-box" or "_log_-driver".

![image](https://user-images.githubusercontent.com/31491646/193116744-13ce337b-7bd8-476a-b6a7-8f1c40d7500e.png)

I spoke with @qtomlinson and it appears this selection behavior may be by design in this view. In the workspace view, if you search and select, only the selected component is added. This should probably be tracked separately in another issue.

bduranc:
I can confirm the above appears resolved in dev environment.

The very last point I am told is by design. However, it does now actually populate the list with what I select (albeit along with many other similarly named components from other providers).

bduranc:
> However, it does now actually populate the list with what I select (albeit along with many other similarly named components from other providers).

![image](https://user-images.githubusercontent.com/31491646/211383536-77334a21-334d-4fb5-b79f-2f399643d49e.png)

@qtomlinson is this by design?

qtomlinson:
@bduranc Yes, what you described in the above screen shot is the designed behaviour in the current web ui.  The original issue documented is that the dropdown list below the search text ('abaus') box did not respect the package type selected ('NPMJS').  The dropdown list now only displays the npm related packages, which addressed the issue of 'search suggestions show everything even though filter is set to "NpmJs"'.
 
[...]

PR #990, opened 2022-07-08
Support component search by provider

qtomlinson:
1. Added provider to the pattern string for coordinates suggestion
The search api call in service ignores provider type.
For example, the following call return co-ordinates with provider other than npmjs:
curl -X GET "https://dev-api.clearlydefined.io/definitions?pattern=jw&type=npm" -H "accept: */*"
This is probably by design: the suggestion api only takes the pattern
string into account.

When provider is included in the pattern string, the search result
reflects the provider specification. The solution is to pass the
provider information when searching in PageBrowse.

2. Also trigger search when provider is changed

Task: https://github.com/clearlydefined/website/issues/957

### DIFF
--- a/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
+++ b/src/components/Navigation/Pages/PageBrowse/PageBrowse.js
@@ -34,7 +34,8 @@ class PageBrowse extends SystemManagedList {
     this.state = {
       activeSort: 'releaseDate-desc',
       searchFocused: false,
-      selectedProvider: providers[0]
+      selectedProvider: providers[0],
+      searchTerm: ''
     }
     this.onFilter = this.onFilter.bind(this)
     this.onSort = this.onSort.bind(this)
@@ -63,11 +64,17 @@ class PageBrowse extends SystemManagedList {
   }
 
   onBrowse = value => {
-    this.setState({ activeName: value }, () => this.updateData())
+    this.setState({ activeName: value, searchTerm: '' }, () => this.updateData())
   }
 
   onFocusChange = value => {
     this.setState({ searchFocused: value })
+  }
+
+  onSearch = value => {
+    this.setState({ searchTerm: value })
+    const provider = this.state.selectedProvider.value
+    super.onSearch(provider + '/' + value)
   }
 
   tableTitle() {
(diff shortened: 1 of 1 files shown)


qtomlinson:
@disulliv @mpcen any feedback?

disulliv (approved):
Not super familiar with this part of the code, but what happens if the provider is empty or we don't want to specify the provider?

edit - sorry about the slow response, it's been a busy few weeks!

qtomlinson:
@disulliv On the UI, searching for components is always within a certain provider. It is initially set to be npm, at PageBrowse (line 37).
image

mpcen:
Can we hold off merging this until searching is fixed. I'm current looking at the search issues

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
```diff
@@ -63,13 +64,19 @@ class PageBrowse extends SystemManagedList {
...
   }
 
   onFocusChange = value => {
     this.setState({ searchFocused: value })
   }
 
+  onSearch = value => {
+    this.setState({ searchTerm: value })
+    const provider = this.state.selectedProvider.value
+    super.onSearch(provider + '/' + value)
```
mpcen, line 77: rather than calling super, can we pass in the method as a prop instead? More on this anti-pattern can be found in [this](https://github.com/facebook/react/issues/3280) discussion

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
```diff
@@ -163,6 +170,7 @@ class PageBrowse extends SystemManagedList {
 
   onProviderChange(item) {
     this.setState({ selectedProvider: item })
+    this.state.searchTerm && super.onSearch(item.value + '/' + this.state.searchTerm)
```
mpcen, line 173: Same feedback from above

src/components/Navigation/Pages/PageBrowse/PageBrowse.js
mpcen: nevermind, this is merged. We'll come back to this

mpcen (commented):
Ignore feedback. PR has already been merged

qtomlinson:
Address feedback in https://github.com/clearlydefined/website/pull/1000

Issue #946, opened 2022-02-14
Issues with Workspace page (new UI)


bduranc:
I found two new issues under the ["Workspace" page](https://clearlydefined.io/workspace) (main/production instance).

- The "Share" dropdown button doesn't show any options when clicked.

![image](https://user-images.githubusercontent.com/31491646/153960367-0b295a20-7163-4866-a536-b0f31fca20b7.png)

- The option to access linked 'source-archive' definitions (important for Maven and Debian components) appears gone and clicking "Add Source Definition" menu item does not appear to fix this.

![image](https://user-images.githubusercontent.com/31491646/153960573-bd171b5a-c7c0-4299-8ccb-8accda795a6e.png)

How do we access the source archive of a Maven or Debian component?

bduranc:
Adding one more problem on the Workspace page:

It appears you can't see items in the list past a certain point (only way is to zoom-out on your browser until visible).

![image](https://user-images.githubusercontent.com/31491646/160506435-828d2040-0af3-4118-b028-1247a2e515aa.png)

It would be nice if we had a scrollbar or something.

bduranc:
I tested the changes in dev environment and can confirm the features now work as they should.

### Test result of "Share" button/feature:

Example 1: https://dev.clearlydefined.io/share/eJyTkuISyyvI1QfirGJ9XRCtb6RnaKxnIsQg5cvlmZ5Zog/EGaVJ+gWVJRn5ebo5VSbIzERDyzRz0xSDNEszM7NE09TUFOPkZFMzQ0MTAwMLwyRzI0Mzy5RUc6BxALOqHu0=

Can confirm populates workspace and produces valid [coordinate.json.txt](https://github.com/clearlydefined/website/files/9677311/coordinate.json.txt)

Example 2: 
https://dev.clearlydefined.io/share/eJxNjMESgiAURV2zdJdN9QX4VArhH1r0Cw8QQxMcoWn6+6xF4+LMWdwztziQnZ8nWBki0K8pzg7qsiqrPCtOZL+dTUedN51Pv6BZgyMptgFGisuC7//BjVx7l2Dl/lQwjHR0vof4cjbpME3BpxAeEWoUpq0apc8MjZS1FRy5YIx3FyOtFVIwrdpW59kHDkk0aA==

Can confirm populates workspace and produces valid [coordinate.json.txt](https://github.com/clearlydefined/website/files/9677392/coordinate.json.txt)

### Test result of "accessing source archive from Worklist":

[...]

PR #998, opened 2022-08-26
Fix scrolling in ComponentList

qtomlinson:
1. The height on the clearly-table-body crops the scroll bar, so
scrolling cannot reach the item at the end of the list.  Moved the
height styling to address this.

2. There is a blank space after components in the table even when
there is not enough space to display all the components.  This is
because renderFilterBar, passed as props from the container components
(e.g. PageDefinition), is not called in ComponentList.  Padding to
accommodate the FilterBar is added by the parent of the ComponentList,
thus leaving a blank space after the ComponentList when FilterBar is
not rendered inside ComponentList.

Change the spacing from padding in the section (parent) to margin of
the ComponentList (child), where showing filter bar is determined.
Also add the conditional formatting to control the spacing based on
whether filter bar is displayed.

Task: https://github.com/clearlydefined/website/issues/946

### DIFF
--- a/src/components/ComponentList.js
+++ b/src/components/ComponentList.js
@@ -145,8 +145,9 @@ class ComponentList extends React.Component {
   render() {
     const { loadMoreRows, noRowsRenderer, list, listLength } = this.props
     const { sortOrder, contentSeq } = this.state
+    const showFilterBar = false
     return (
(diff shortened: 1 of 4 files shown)


qtomlinson:
@mpcen ready for review

mpcen (approved):

PR #1001, opened 2022-09-30
fixed broken links

ACAnyasodor:
i don't have npm installed on my pc so i couldn't test the code but with my knowledge of HTML i'm sure the links work just fine. and even if it doesn't, give me feedback.


jeffwilcox:
Thanks for this contribution.

Since the site itself is built using the React framework, at build time these links then get deployed and working. At this time, closing this.

PR #1038, opened 2024-04-04
Remove Azure Pipelines config

lumaxis:
We now have a GitHub Actions workflow (#1035), so this will close https://github.com/clearlydefined/crawler/issues/550

Someone with Admin permissions to this repo will need to remove the Pipelines job as a required job and add the Actions build instead.


elrayle:
I disabled the Azure DevOps auto-deploy based on the master branch trigger.

.github/workflows/test.yml
```diff
@@ -30,3 +30,6 @@ jobs:
 
       - name: Run tests
         run: npm test
+
+      - name: Docker build
+        run: docker build -t clearlydefined/website:latest .
```
elrayle, line 35: My take on building the image as part of the test workflow is that it confirms that the PR running the tests did not break the ability to build the docker image.  This is a useful block to prevent merge of a PR that will break the docker build that happens on merge to master branch.  Is that the intent?

lumaxis: It was part of the previous Azure DevOps build in the same place and that is my best guess of its intention, yes

PR #1052, opened 2024-06-03
docs: change footer

nickvidal:
Change the footer description, as ClearlyDefined is not an incubator project anymore. We can add reference to Microsoft's donation of the project in the documentation.


elrayle (approved):
Thanks for keeping docs up to date.  The footer is very prominent and good to have it showing CD in the best light.

PR #1060, opened 2024-07-24
Remove quotes from all docker build-args

elrayle:
### Description

Build args that are in quotes have the quotes encoded leading to an incorrect value when the Docker image is built.  This removes the quotes.  The dev website was confirmed to load correctly after this change.


qtomlinson (approved):
Thanks for the fix.  LGTM

PR #1068, opened 2024-10-30
Add health endpoint via nginx

ljones140:
Adds a health endpoint which returns app version and sha.

```
{"status":"OK", "version": "2.2.1", "sha": "12345"}
```

Endpoint has been implemented via nginx.
As the site is a React page the alternative would have been adding a Node.js endpoint the app would need to expose.
Using nginx is a lot less complex and easier to maintain in the future

Deployed to development and it's working:
```
curl https://dev.clearlydefined.io/health
=> {"status":"OK", "version": "v0.1.1-dev-654d7fca75", "sha": "654d7fca7516735b2e9891b8657352cf3b3e0a4e"}
```

### DIFF
--- a/.github/workflows/build-and-deploy-dev.yml
+++ b/.github/workflows/build-and-deploy-dev.yml
@@ -33,7 +33,7 @@ jobs:
   build-and-deploy:
     name: Build and Deploy
     needs: [upload-package-lock-json, make-react-secret-available]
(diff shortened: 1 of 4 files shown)


elrayle (approved):
Fantastic.  I'm glad you found an easy solution to make the health end point.

PR #1118, opened 2026-03-20
Adding exception picker to license picker form

capfei:
This allows UI users to submit curations with SPDX exceptions (WITH). More fixes to correct the table layout.

Test PRs submitted: 

- https://github.com/clearlydefined/curated-data-dev/pull/1640
- https://github.com/clearlydefined/curated-data-dev/pull/1641

New WITH button:
image

New exception picker after clicking button:
image

Exception picker list:
image

Table layout fix:
image


ashleywolf:
Thanks for adding SPDX exception support. A few things that came up to consider before merging:

**Data corruption: space sentinel in exception initialization**

Clicking "+ WITH" calls `updateException(' ', path)`, setting the exception to a single space. Since `' '` is truthy, `LicensePickerUtils.toString()` serializes this as `"MIT WITH "` -- an invalid SPDX expression. The `.trim()` in the RuleBuilder display only affects the picker input, not the stored value. If a user clicks "+ WITH" and saves without selecting an exception, invalid data gets persisted.

Suggestion: use `null` or a separate boolean flag to track "picker is open" state instead of a whitespace sentinel. Or trim the exception in `toString()` and treat empty/whitespace as absent.

**Shallow copy + nested mutation in updateException**

`updateException` does `{ ...this.state.rules }` (shallow copy) then mutates the nested object directly via `currentRule.exception = value`. For nested rules (inside `combinations`), `get(shallowCopy, path)` returns a reference to the original state object, so this mutates `this.state.rules` directly. Same issue exists in the new `updateLicense` cleanup path. Could cause missed re-renders.

`updateLicense` already uses `lodash/set` with `toPath` for the license value -- same pattern would work here.

**Clearing a license doesn't clear the `plus` flag**

The new code clears `exception` when the license is cleared, but leaves `plus` intact. If a user checks "Any later version", removes the license via the new root-level remove button, then picks a different license, the old `+` silently carries over (e.g., new license becomes `MIT+` without the user re-enabling it).

**Minor items:**

- `spdxExceptions.sort()` in the constructor mutates the imported module array in-place. Use `[...spdxExceptions].sort()` to sort a copy.
- `defaultInputValue` in `SpdxExceptionPicker` is uncontrolled (initial-only). If the parent re-renders with a new value, the input won't update. Consider `key={value}` to force remount, or a controlled `inputValue` prop if the Autocomplete supports it.
- `index.css.map` is a generated artifact -- should be in `.gitignore`, not committed.
- The Netlify deploy preview is failing. Could be related to the `index.css` -> `index.scss` import change in `src/index.js`?
- Worth confirming the FileList changes are intentional -- removing `expandedRowKeys` and setting `childrenColumnName="_nochildren"` disables tree expansion entirely, which means search results inside subdirectories won't render even though `filterFiles()` still returns nested `children`.

capfei:
@ashleywolf Thank you for the feedback! ❤️ I'll start looking at each of the issues mentioned. I didn't even think to check the `+` checkbox.🤦‍♀️

